PR #35, opened 2025-09-17
Handle trailing zeros in AlignedValue

jsidorenko:


integration-tests/src/test/no-proving/functions.test.ts
```diff
@@ -726,6 +726,67 @@ describe('Ledger API - functions', () => {
...
+        { tag: 'atom', value: { tag: 'bytes', length: 1 } },
+        { tag: 'atom', value: { tag: 'bytes', length: 32 } },
+        { tag: 'atom', value: { tag: 'bytes', length: 32 } }
+      ]
+    };
+    const commitment = runtimeCoinCommitment(coin, recipient);
+    expect(Buffer.from(commitment.value[0]).toString('hex')).toEqual(
+      'e5b99c814053184ad9d37d03fc4558d03982c418681561d1f0c2c76bd4a624'
+    );
+    expect(commitment.value[0].length).toEqual(PERSISTENT_HASH_BYTES);
```
tkerber, line 758: As discussed on slack, this is *not* a correct assumption.

tkerber (changes requested):
I do not consider either of these tests to be correct or to be bugs in `AlignedValue`. They are usage bugs; `AlignedValue` is an encoding that does not allow for trailing zeros.

PR #37, opened 2025-09-18
Correctly trim trailing zeros in manually constructed aligned values.

tkerber:
This seems to step from a misunderstanding of what the `encode*` functions do – they do _not_ target `[Aligned]Value`, but rather the Compact-native encoding.

### DIFF
--- a/integration-tests/src/test/utils/RuntimeCoinCommitmentUtils.ts
+++ b/integration-tests/src/test/utils/RuntimeCoinCommitmentUtils.ts
@@ -43,7 +43,11 @@ export class RuntimeCoinCommitmentUtils {
     const isUserAddress = false;
 
     return {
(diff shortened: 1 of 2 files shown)


jacek-kurkowski-shielded (approved):

integration-tests/src/test/utils/RuntimeCoinCommitmentUtils.ts
```diff
@@ -65,13 +69,14 @@ export class RuntimeCoinCommitmentUtils {
   }
 
   static assertOutcomes(coin: AlignedValue, recipient: AlignedValue) {
+    console.log(JSON.stringify(coin, null, 4), JSON.stringify(recipient, null, 4));
```
jsidorenko, line 72: pls remove that

PR #44, opened 2025-09-22
Expose functions for maximum price adjustment / safety margin for fees

tkerber:
Takes a bit of ownership of how much fees transactions _should_ pay, by exposing a margin that covers maximum upward adjustment for `n` blocks.

### DIFF
--- a/.github/workflows/prepare-cost-model-docker.yml
+++ b/.github/workflows/prepare-cost-model-docker.yml
@@ -72,7 +72,7 @@ jobs:
         run: .github/scripts/configure-cargo.sh
 
       - name: Install Rust Toolchain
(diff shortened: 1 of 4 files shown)


ledger-wasm/src/tx.rs
```diff
@@ -693,6 +693,10 @@ impl Transaction {
         get_dyn_transaction(self.0.clone()).fees(params)
     }
 
+    pub fn fees_with_margin(&self, params: &LedgerParameters, n: usize) -> Result<BigInt, JsError> {
```
jacek-kurkowski-shielded, line 696: Don't you need: `#[wasm_bindgen(js_name = "feesWithMargin")]` at the top?

PR #63, opened 2025-10-10
test: proof-server adding integration tests

yosuancrespo:
Added integration tests:

- Malformed format – no ZK Config
test_prove_should_fail_without_zk_config

- Malformed format – double ZK Config
test_prove_should_fail_with_double_zk_config

- Malformed format – two ZK Configs (swapped values)
test_prove_should_fail_with_swapped_zk_config_values

- Malformed format – wrong circuitId
test_prove_should_fail_with_wrong_circuit_id

- Generate valid proof – smallest size transaction
test_prove_should_generate_valid_proof_for_smallest_transaction

- Generate valid proof – big transaction
test_prove_should_generate_valid_proof_for_big_transaction

- Mismatch Network ID
test_prove_should_fail_with_mismatched_network_id

**Note:**

In this PR, the pre-existing tests were migrated to use the /prove endpoint instead of the deprecated /prove-tx endpoint.


jacek-kurkowski-shielded (changes requested):
Hi @yosuancrespo !

Happy to discuss the comments if anything is unclear.

integration-tests/src/test/proving/ProveTxNoBody.test.ts
```diff
@@ -0,0 +1,33 @@
+// This file is part of midnight-ledger.
```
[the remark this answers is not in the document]
yosuancrespo: The test for sending POST to /prove-tx without body is:
test_prove_tx_should_fail_on_empty_body()

proof-server/tests/integration_tests.rs
```diff
@@ -108,18 +114,159 @@ async fn stop_server(server_handle: ServerHandle) {
...
+        ) = tagged_deserialize(&serialized_valid_zswap_body().await[..]).ok()?;
+        let mut existing_keys: HashSet<_> = entries.iter().map(|(key, _)| key.clone()).collect();
+        entries.extend(
+            extra_keys
+                .into_iter()
+                .filter(|(key, _)| existing_keys.insert(key.clone())),
+        );
+    }
+
+    if entries.len() < 2 {
```
jacek-kurkowski-shielded: Hmm, I don't get this check 😄 
Isn't the case of the code above to have 2 keys in entries?

proof-server/tests/integration_tests.rs
```diff
@@ -108,18 +114,159 @@ async fn stop_server(server_handle: ServerHandle) {
...
+    let valid_body = serialized_valid_body().await;
+    let (tx, keys): (
+        Transaction<Signature, ProofPreimageMarker, PedersenRandomness, InMemoryDB>,
+        std::collections::HashMap<
+            transient_crypto::proofs::KeyLocation,
+            transient_crypto::proofs::ProvingKeyMaterial,
+        >,
+    ) = tagged_deserialize(&valid_body[..]).ok()?;
+
+    let mut entries: Vec<_> = keys.into_iter().collect();
```
jacek-kurkowski-shielded: Is it needed at all? Can't you use `keys` directly?

proof-server/tests/integration_tests.rs
```diff
@@ -108,18 +114,159 @@ async fn stop_server(server_handle: ServerHandle) {
...
```
jacek-kurkowski-shielded: As mentioned earlier.

proof-server/tests/integration_tests.rs
```diff
@@ -108,18 +114,159 @@ async fn stop_server(server_handle: ServerHandle) {
...
+    if entries.is_empty() {
+        let (_, fallback_keys): (
+            Transaction<Signature, ProofPreimageMarker, PedersenRandomness, InMemoryDB>,
+            std::collections::HashMap<
+                transient_crypto::proofs::KeyLocation,
+                transient_crypto::proofs::ProvingKeyMaterial,
```
jacek-kurkowski-shielded: Same here.

proof-server/tests/integration_tests.rs
```diff
@@ -339,6 +559,216 @@ async fn test_prove_tx_should_fail_on_json() {
...
+// Negative test: `/prove` must reject payloads whose ZK Config contains an
+// entry keyed by a circuit ID that is not part of the transaction.
+// Given: A valid request `(tx, {A -> zkA, ...})`.
+// When: We replace one key with `C_wrong` while keeping the proving material.
+// Then: The server responds with `400 Bad Request`.
+#[named]
+async fn test_prove_should_fail_with_wrong_circuit_id() {
+    setup_test(function_name!());
+
+    if let Some(payload) = serialized_body_with_wrong_circuit_id().await {
```
jacek-kurkowski-shielded: This will silently pass the test if `serialized_body_with_swapped_zk_config_values` fails 😢

jacek-kurkowski-shielded (changes requested):
Hi @yosuancrespo 
I just left a few comments. Feel free to reach out if anything is unclear.

proof-server/tests/integration_tests.rs
```diff
@@ -204,13 +408,29 @@ async fn integration_tests() {
     test_prove_tx_should_fail_on_get().await;
     test_prove_tx_should_fail_on_empty_body().await;
     test_prove_tx_should_fail_on_json().await;
+    test_prove_should_fail_without_zk_config().await;
+    test_prove_should_fail_with_double_zk_config().await;
+    test_prove_should_fail_with_swapped_zk_config_values().await;
+    test_prove_should_fail_with_wrong_circuit_id().await;
+    test_prove_should_generate_valid_proof_for_smallest_transaction().await;
+    test_prove_should_generate_valid_proof_for_big_transaction().await;
```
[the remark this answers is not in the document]
yosuancrespo: I created this ticket: https://shielded.atlassian.net/browse/PM-20202. Please let me know if anything needs to be modified.

proof-server/tests/integration_tests.rs
yosuancrespo: Great point. Updated the helper to operate on the resolver map directly, so we reuse the keys in place.

proof-server/tests/integration_tests.rs
```diff
@@ -339,6 +559,216 @@ async fn test_prove_tx_should_fail_on_json() {
...
+
+    let response = HTTP_CLIENT
+        .post(format!("{}/prove", get_host_and_port()))
+        .body(serialized_body_without_zk_config().await)
+        .send()
+        .await
+        .unwrap();
+
+    log::info!("Response code: {:?}", response.status());
+    assert_eq!(response.status(), 400);
```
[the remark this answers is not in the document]
yosuancrespo: Thanks for the pointer, these tests now pull in `reqwest::StatusCode`, and assert against `StatusCode::BAD_REQUEST` instead of raw integers

proof-server/tests/integration_tests.rs
```diff
@@ -108,18 +114,159 @@ async fn stop_server(server_handle: ServerHandle) {
...
 async fn serialized_valid_zswap_body() -> Vec<u8> {
     let tx = valid_unbalanced_zswap(1);
-    #[allow(deprecated)]
-    serialize_request_body(&tx, &RESOLVER).await.unwrap()
+    serialize_transaction_payload(&tx, &RESOLVER).await
+}
+
+// Serializes a given **unproven transaction (preimage)** plus resolver-provided proving keys
+// into the exact tagged payload that `/prove` expects.
+async fn serialize_transaction_payload<S, D>(
```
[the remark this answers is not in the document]
yosuancrespo: Good catch! I’ve pinned the helpers to `InMemoryDB`, dropped the extra D generic, and adjusted the calls accordingly.

proof-server/tests/integration_tests.rs
yosuancrespo: Good point, this helper only needs to bail when the resolver can’t give us two locations. I removed that explicit guard.

proof-server/tests/integration_tests.rs
```diff
@@ -184,15 +337,66 @@ async fn valid_tx<S: SignatureKind<D>, D: DB>()
...
+    unsafe {
+        env::set_var("NETWORK_ID", value);
+    }
+}
+
+fn clear_network_id() {
+    unsafe {
+        env::remove_var("NETWORK_ID");
+    }
+}
```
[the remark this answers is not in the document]
yosuancrespo: `env::set_var` and `env::remove_var` are marked unsafe in our current toolchain, that's why the calls are inside an `unsafe` block. I included comments on why that usage is sound: tests run the proof server on a dedicated thread and no other code mutates the process environment concurrently. Below you can see the output if we don't use the `unsafe` blocks. If you know a workaround for this I am happy to work on that.

```
error[E0133]: call to unsafe function `set_var` is unsafe and requires unsafe block
   --> proof-server/tests/integration_tests.rs:356:5
    |
356 |     env::set_var("NETWORK_ID", value);
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ call to unsafe function
    |
    = note: consult the function's documentation for information on how to avoid undefined behavior

error[E0133]: call to unsafe function `remove_var` is unsafe and requires unsafe block
   --> proof-server/tests/integration_tests.rs:360:5
    |
360 |     env::remove_var("NETWORK_ID");
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ call to unsafe function
    |
    = note: consult the function's documentation for information on how to avoid undefined behavior
```

proof-server/tests/integration_tests.rs
```diff
@@ -339,6 +559,216 @@ async fn test_prove_tx_should_fail_on_json() {
...
+// Given:
+// - A valid request `(tx, {A -> zkA, B -> zkB, ...})`.
+// - We swap the proving material so that `{A -> zkB, B -> zkA}`.
+// When: The malformed payload is POSTed to `/prove`.
+// Then: The server responds with `400 Bad Request`.
+#[named]
+async fn test_prove_should_fail_with_swapped_zk_config_values() {
+    setup_test(function_name!());
+
+    if let Some(payload) = serialized_body_with_swapped_zk_config_values().await {
```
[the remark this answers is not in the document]
yosuancrespo: Added `.expect(...)` when constructing the payload so the test aborts immediately if `serialized_invalid_body_with_swapped_zk_config_values()` fails, addressing the silent-pass concern.

jacek-kurkowski-shielded:
Not relevant anymore

PR #66, opened 2025-10-13
Update hard fork patch script

NachoPal:
Update hard fork patch script.

Partially closes: https://shielded.atlassian.net/browse/PM-19841


gilescope:
are we no longer formatting the code?

PR #71, opened 2025-10-17
`storage`: Embed Small Nodes

AlexShielded:
Opened for initial review due to going on annual leave.

Tests still failing:
* `storage::tests::persist_to_disk_sqldb` (but strangely `parity_db` works...)

### DIFF
--- a/storage/src/arena.rs
+++ b/storage/src/arena.rs
@@ -840,6 +840,13 @@ impl<D: DB> Arena<D> {
             result = Ok(root);
         }
 
+        // TODO: Suggested fix for missing ArenaKey representation: do a graph-walk bottom-up on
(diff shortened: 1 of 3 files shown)


tkerber:
Some initial notes working with this (Edit: I'd accidentally not pulled in the most recent commit initially. Updated thoughts):
- Worryingly, there are warnings that `CacheValue::CreateAndDelete`, `StorageBackend::uncache` and `Cache::update_in_place` are unused. I believe I did disable these initially, and it seems they still haven't been properly re-enabled.
- `Storage::get` and `Storage::get_lazy` should almost definitely now take `ChildNode`. We probably want a typed variant of this as well for the type-safety aspect.

PR #77, opened 2025-10-23
[WASM] Standartize toString() usage in Wasm

jsidorenko:
In toString() methods we should print the internal object without a wrapper:
```
 - format!("{:?}", &self)
 + format!("{:?}", &self.0)
```
Additionally, adds missing getter tags in DustState


jacek-kurkowski-shielded (approved):
Love it, thanks! 😄

PR #78, opened 2025-10-23
Fix: solve reported clippy issue for rule type_complexity

bochaco:
This solves a few issues reported by clippy with regards to [type_complexity](https://rust-lang.github.io/rust-clippy/master/index.html#type_complexity) rule.

### DIFF
--- a/ledger/src/error.rs
+++ b/ledger/src/error.rs
@@ -691,18 +691,18 @@ pub struct SubsetCheckFailure<T> {
 /// The inner tuple holds the contract address, a hash output (such as a commitment or nullifier),
 /// and a field element (Fr), which may represent a value or cryptographic proof.
 /// This structure is commonly used to identify and track contract calls within a specific segment.
-pub type SegmentContractCall = (u16, (ContractAddress, HashOutput, Fr));
+pub type ContractCallReference = (u16, (ContractAddress, HashOutput, Fr));
 
 #[derive(Clone, Debug)]
 pub enum EffectsCheckError {
-    RealCallsSubsetCheckFailure(SubsetCheckFailure<SegmentContractCall>),
+    RealCallsSubsetCheckFailure(SubsetCheckFailure<ContractCallReference>),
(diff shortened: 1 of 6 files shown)


ledger/src/prove.rs
```diff
@@ -86,19 +87,20 @@ impl transient_crypto::proofs::Resolver for Resolver {
     }
 }
 
+// Map of prooved intents
+type ProovedIntents<S, D> =
```
tkerber: Spelling nit: This should be `Proved`

ledger/src/structure.rs
```diff
@@ -759,6 +759,9 @@ pub struct Intent<S: SignatureKind<D>, P: ProofKind<D>, B: Storable<D>, D: DB> {
 }
 tag_enforcement_test!(Intent<(), (), Pedersen, InMemoryDB>);
 
+// Intent with a PedersenRandomness
+pub(crate) type PedersenRandomnessIntent<S, P, D> = Intent<S, P, PedersenRandomness, D>;
```
tkerber: I'm not really a fan of this. It doesn't actually simplify the type, just shuffles around the wording and introduces indirection. I believe in most cases here we also use `P = ProofPreimageMarker`, so maybe an alias like `type UnprovenIntent<S, D> = Intent<S, ProofPreimageMarker, PedersenRandomness, D>` makes more sense?

ledger/src/error.rs
```diff
@@ -687,16 +687,22 @@ pub struct SubsetCheckFailure<T> {
     pub subset: Vec<T>,
 }
 
+/// Type alias for a tuple containing a segment ID and contract call details.
+/// The inner tuple holds the contract address, a hash output (such as a commitment or nullifier),
+/// and a field element (Fr), which may represent a value or cryptographic proof.
+/// This structure is commonly used to identify and track contract calls within a specific segment.
+pub type SegmentContractCall = (u16, (ContractAddress, HashOutput, Fr));
```
tkerber: Suggest calling this `ContractCallReference`

PR #91, opened 2025-11-14
ZKIR v3 Pt 1

JosephDenman:
Adds support for ZKIR v3.

1. Implements the ZKIR v3 interpreter with support for named inputs and outputs for ZKIR v3 instructions. Also has support for constant operands.
2. Implements a wrapper around ZKIR v2 and ZKIR v3 which delegates to either one depending on the version string provided in the ZKIR JSON.
3. Updates `zkir-wasm` to handle both versions.

### DIFF
--- a/.tag-decompositions/prover-key[v6](ir-source[v3])
+++ b/.tag-decompositions/prover-key[v6](ir-source[v3])
@@ -1,0 +1,1 @@
+prover-key[v6](ir-source[v3])
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -481,7 +481,7 @@ checksum = "55248b47b0caf0546f7988906588779981c43bb1bc9d0c44087278f80cdb44ba"
 [[package]]
 name = "bench_macros"
 version = "0.1.0"
-source = "git+https://github.com/midnightntwrk/midnight-zk?tag=midnight-circuits-v5.0.1#54bd684da90b4c7a24cb9a40f447e6c06ee40a7a"
+source = "git+https://github.com/midnightntwrk/midnight-zk?tag=midnight-circuits-v5.0.2#64ba405fedab8c96e09abde3dc134c902c6f2c6f"
 dependencies = [
  "quote 1.0.42",
  "syn 2.0.110",
@@ -1814,7 +1814,7 @@ dependencies = [
 
 [[package]]
 name = "generate-cost-model"
-version = "1.0.0-rc.1"
+version = "1.0.0-rc.2"
 dependencies = [
  "charming",
  "clap",
@@ -2714,7 +2714,7 @@ dependencies = [
 
 [[package]]
 name = "midnight-base-crypto"
-version = "1.0.0-rc.1"
+version = "1.0.0-rc.2"
 dependencies = [
  "actix-rt",
  "anyhow",
@@ -2750,7 +2750,7 @@ dependencies = [
 
 [[package]]
 name = "midnight-base-crypto-derive"
-version = "1.0.0-rc.1"
+version = "1.0.0-rc.2"
 dependencies = [
  "proc-macro2",
  "quote 1.0.42",
@@ -2759,8 +2759,8 @@ dependencies = [
 
 [[package]]
 name = "midnight-circuits"
-version = "5.0.1"
-source = "git+https://github.com/midnightntwrk/midnight-zk?tag=midnight-circuits-v5.0.1#54bd684da90b4c7a24cb9a40f447e6c06ee40a7a"
+version = "5.0.2"
+source = "git+https://github.com/midnightntwrk/midnight-zk?tag=midnight-circuits-v5.0.2#64ba405fedab8c96e09abde3dc134c902c6f2c6f"
 dependencies = [
  "base64 0.13.1",
  "bench_macros",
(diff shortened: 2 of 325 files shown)


zkir/src/v3/ir.rs
```diff
@@ -0,0 +1,643 @@
...
+        count: u32,
+    },
+    /// Adds two elliptic curve points. UB if either is not a valid curve point.
+    ///
+    /// Outputs 2 elements, `c_x`, `c_y`
+    EcAdd {
+        /// The affine x coordinate of `a`
+        a_x: Operand,
+        /// The affine y coordinate of `a`
+        a_y: Operand,
```
tkerber: Wasn't typed values in ZKIR one of the main things for v3? Where this should be and `a` and `b` would be the operands here.

tkerber:
> This isn't a full implementation of ZKIR v3.

That's good to know. It also strengthens the case for this being in a separate package, because that'll keep it out of the stable dependency graph, and have future ledger releases pull in a work-in-progress ZKIR release.

zkir/src/lib.rs
```diff
@@ -16,19 +16,19 @@ extern crate tracing;
...
-use serialize::tagged_deserialize;
 use transient_crypto::curve::Fr;
 use transient_crypto::proofs::{
     ParamsProverProvider, Proof, ProofPreimage, ProvingProvider, Resolver,
 };
 
-mod ir;
-mod ir_vm;
+pub mod v2;
+pub mod v3;
```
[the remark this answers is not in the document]
JosephDenman: Done

zkir-v3-wasm/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+    );
+    let padded_str = if hex_str.len() % 2 == 1 {
+        "0".to_owned() + &hex_str
+    } else {
+        hex_str
+    };
+    let mut bytes = <Vec<u8>>::from_hex(padded_str.as_bytes())?;
+    bytes.reverse();
+    Fr::from_le_bytes(&bytes).ok_or_else(|| JsError::new("out of bounds for prime field"))
+}
```
miguel-ambrona: Shall this function be in another file, devoted to type conversions?

Do we want to fail if the BigUint is bigger than the field order? Shall we document that?

In `midnight-zk` we implement a similar function in a simpler and more readable way (without string conversions).

```rust
    let (sign, u64_chunks) = bigint.to_u64_digits();
    let res = from_u64_le_digits::<Fr>(&u64_chunks);
    if sign == Sign::Minus { -res } else { res }
```

Where:

```rust
fn from_u64_le_digits<F: PrimeField>(digits: &[u64]) -> F {
    if digits.is_empty() {
        return F::ZERO;
    }

    let mut acc = F::from(*digits.last().unwrap());
    for digit in digits.iter().rev().skip(1) {
        for _ in 0..64 {
            acc = acc.double();
        }
        acc += F::from(*digit)
    }
    acc
}
```

miguel-ambrona (approved):
LGTM.

I understand you want to update and test Compact incrementally. Starting the v3 VM as (almost) a copy of v2 makes sense for that, but I am afraid this will make the transition to a typed IR harder than if we redesigned v3 from scratch (as I did [here](https://github.com/midnightntwrk/midnight-zk/tree/zkir-v3-new-ops-3/zkir)).

I noticed, in this repository most functions lack documentation.

This PR is not the place to do it, but some things that come to mind, which we should do IMHO:
 - Get rid of the communications commitment in ZKIR: our IR is rich enough to encode this via instructions. It could be Compact who, though a communications commitment feature, produced the necessary IR instructions to handle it.
- Get rid of pi_skips (I believe this is already part of the plan).
- Organize serialization of constants in a new file (we only have field elements now, but we will have more types).
- Try to adopt the set of instructions that I detailed [here](https://github.com/midnightntwrk/midnight-zk/blob/zkir-v3-new-ops-3/zkir/src/instructions/operations/mod.rs#L9).

proof-server/src/lib.rs
```diff
@@ -163,12 +164,20 @@ async fn proof_versions() -> impl Responder {
...
     );
+
+    // Version detection - try V2 first, then V3
+    let k = if let Ok(ir_v2) = tagged_deserialize::<zkir_v2::IrSource>(&request[..]) {
+        ir_v2.k()
+    } else if let Ok(ir_v3) = tagged_deserialize::<zkir_v3::IrSource>(&request[..]) {
+        ir_v3.k()
+    } else {
+        return Err(ErrorBadRequest("Unsupported ZKIR version"));
+    };
```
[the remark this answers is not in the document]
MicroProofs: Is the tag header 0 index based. I know rust has not straightforward rules for where the bits line up with enums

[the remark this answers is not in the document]
MicroProofs: I can change this in a bit. just wondering which direction we want to go

proof-server/Cargo.toml
```diff
@@ -6,6 +6,7 @@ version.workspace = true
 
 [features]
 default = []
+experimental = []
```
tkerber:
```suggestion
experimental = ["dep:zkir-v3"]
```

serialize/src/deserializable.rs
```diff
@@ -52,6 +52,7 @@ pub fn tagged_deserialize<T: Deserializable + Tagged>(mut reader: impl Read) ->
     }
     let value = <T as Deserializable>::deserialize(&mut reader, 0)?;
 
+    #[allow(clippy::unbuffered_bytes)]
```
tkerber: Ditto, please remove.

PR #79, opened 2025-10-27
Audit fixes

tkerber:
Fix audit issues identified by Least Authority.

PR #101, opened 2025-12-01
Change pricing structure to use an overall cost and dimension weightings

tkerber:
This is necessary because:
1. The node does some work that is not broken down into ledger cost dimensions
2. The node's transaction selection is not aware of cost dimensions, and this could lead to a mismatch of expectations, where the ledger consider's a block mostly empty, but the node considers it full, leading to a downward price spiral.

Stacked behind #79


integration-tests/src/test/no-proving/CostModel.test.ts
```diff
@@ -55,14 +56,17 @@ describe('Ledger API - CostModel', () => {
    * block. The prices you read now apply to the *next* block.
    * 2. Each dimension increases if its utilization exceeds the target,
    * decreases if below, and remains steady near the target.
-   * 3. Price movement per block is multiplicatively bounded by `maxPriceAdjustment()` (~1.045). This cap limits how fast prices
+   * 3. Overall price is adjusted based on the overall fullness:
+   *    - Price movement per block is multiplicatively bounded by `maxPriceAdjustment()` (~1.045). This cap limits how fast prices
```
jacek-kurkowski-shielded: This line is cut. Later there's `...can change upward or downward between consecutive blocks.`

kapke (approved):

PR #109, opened 2025-12-05
Fix: solve reported clippy issue for rule too_many_arguments

bochaco:
This solves a few issues reported by clippy with regards to [too_many_arguments](https://rust-lang.github.io/rust-clippy/master/index.html#too_many_arguments) rule.

BREAKING CHANGE:
Change in `Intent::new` method/API of the `construct` module to now receive an argument (defined as type `IntentParams<S, D>`) which contains all attributes to be used for creating the `Intent`, instead of expecting all those attributes as separate function arguments.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -218,8 +218,8 @@ jobs:
       # We still permit certain Clippy rules with `-A` args. since they will be
       # addressed in upcoming PRs and changes that are currently in progress. This
       # allows us to start running these checks even before they are fully implemented.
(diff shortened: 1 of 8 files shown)


tkerber:
I was just reviewing the list of PRs, and realised I never wrote up a response to this.

I don't think I can accept this PR as-is, because a) it's a pretty invasive breaking change to the API, and b) it replaces constructor functions with plain objects that are used to initialize more structured objects. That feels like an anti-pattern to me.

At this stage, I'm more inclined to accept the places where clippy complains about argument count.

PR #115, opened 2025-12-08
[NOT TO BE MERGED] debug: Add comprehensive logging for dust spend proof verification (block 8226 investigation)

cosmir17:


cosmir17:
Closing - the block 8226 issue has been resolved via a different approach.
Root cause: Transaction ordering bug in the indexer (system TXs from events were appended instead of prepended).
Actual fix: https://github.com/midnightntwrk/midnight-indexer/pull/592
This investigation PR is no longer needed.

PR #116, opened 2025-12-09
[PM-20455] Return input/output for dust/zswap state

iangregsondev:
Wallet needs to provide shielded TX history. With the current implementation, there is no efficient way of doing this.

It was suggested that we extend replay_events for the localState of both Zswap and Dust to return input/outputs spent.

Although to avoid breaking changes, we have introduced a new API.


ledger-wasm/src/zswap_state.rs
```diff
@@ -87,6 +88,29 @@ impl MerkleTreeCollapsedUpdate {
...
+#[wasm_bindgen]
+impl ZswapLocalStateWithChanges {
+    #[wasm_bindgen(getter)]
+    pub fn state(&self) -> ZswapLocalState {
+        ZswapLocalState(self.inner.result.clone())
+    }
+
+    #[wasm_bindgen(getter)]
+    pub fn changes(&self) -> Vec<ZswapStateChanges> {
+        self.inner
```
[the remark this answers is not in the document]
iangregsondev: I implemented the From trait for ZswapLocalStateWithChanges and also DustLocalStateWithChanges

ledger/src/dust.rs
```diff
@@ -1322,6 +1323,71 @@ pub struct DustLocalState<D: DB> {
...
+        WithDustStateChanges {
+            changes: Vec::new(),
+            result,
+        }
+    }
+}
+
+impl<T> WithDustStateChanges<T> {
+    pub fn add_change(mut self, change: DustStateChanges) -> Self {
+        if let Some(last_change) = self.changes.first_mut() {
```
kapke: it should be `last_mut`

ledger-wasm/src/state_changes.rs
```diff
@@ -0,0 +1,100 @@
...
+    #[wasm_bindgen(getter)]
+    pub fn source(&self) -> Result<String, JsError> {
+        to_hex_ser(&self.inner.source)
+    }
+
+    #[wasm_bindgen(getter, js_name = "receivedCoins")]
+    pub fn received_coins(&self) -> Result<Array, JsError> {
+        let coins = Array::new();
+        for coin in &self.inner.received_coins {
+            coins.push(&qualified_shielded_coininfo_to_value(coin)?);
```
kapke: Minor, I would prefer https://docs.rs/js-sys/latest/js_sys/struct.Array.html#impl-FromIterator%3CA%3E-for-Array

kapke (approved):
Looks good to me!

ledger-wasm/src/state_changes.rs
iangregsondev: Added this

```
    pub fn received_coins(&self) -> Result<Array, JsError> {
        let coins = self
            .inner
            .received_coins
            .iter()
            .map(qualified_shielded_coininfo_to_value)
            .collect::<Result<Vec<_>, _>>()?;

        Ok(coins.into_iter().collect::<Array>())
    }
```

I am also updating other methods in this file that can take advantage of it.

PR #117, opened 2025-12-11
Log state transitions & script to identify log commonality/divergence

tkerber:
Adds debug logs that print each state transition the ledger does, and a script that compares two such logs to identify where things diverge.

### DIFF
--- a/proof-server/src/endpoints.rs
+++ b/proof-server/src/endpoints.rs
@@ -1,3 +1,16 @@
+// This file is part of midnight-ledger.
+// Copyright (C) 2025 Midnight Foundation
+// SPDX-License-Identifier: Apache-2.0
+// Licensed under the Apache License, Version 2.0 (the "License");
+// You may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// http://www.apache.org/licenses/LICENSE-2.0
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


scripts/state-divergence-identifier.py
```diff
@@ -0,0 +1,90 @@
...
+    print("Usage: scripts/state-divergence-identifier.py <LOGFILE 1> <LOGFILE 2>")
+    print("")
+    print("This script finds a point where the state evolutions of two logfiles diverge.")
+    print("It relies on the ledger 'state transition: <xyz> => <abc>' debug logs,")
+    print("building a tree graph for both, finding the longest path in both, and")
+    print("finding the point these paths diverge")
+else:
+    print("parsing log 1...");
+    graph1 = build_graph(sys.argv[1])
+    print("parsing log 2...");
```
jacek-kurkowski-shielded: Rust coder allert - `;` are not needed 😄

scripts/state-divergence-identifier.py
```diff
@@ -0,0 +1,90 @@
...
+        for child in children:
+            parents[child].append(node)
+    genesis_points = [key for key in graph.keys() if len(parents[key]) == 0]
+    dists = defaultdict(lambda: (math.inf, None))
+    for genesis in genesis_points:
+        dists[genesis] = (0, None)
+    frontier = genesis_points
+    while len(frontier) != 0:
+        curr = frontier.pop()
+        next = graph[curr]
```
jacek-kurkowski-shielded: Change this name as it shadows built-in name 'next'

PR #121, opened 2025-12-17
[PM-17753] Transcript partitioning with zswap components via `addCalls`

tkerber:
As an added convenience, some of the mechanisms I needed here involve modifying zswap inputs/outputs, which means its now possible to create offers without knowing the segment they target, or re-stating the value (and potentially getting it wrong) in deltas.

While I was at it, we now recompute the binding randomness for transactions where possible on modifications in wasm, which is more robust to transaction edits.


ledger-wasm/src/tx.rs
```diff
@@ -853,6 +911,92 @@ impl Transaction {
...
+                    ttl,
+                    &zswap_inputs,
+                    &zswap_outputs,
+                    &zswap_transient,
+                )?;
+                Ok(Transaction(UnprovenWithSignatureErasedPreBinding(
+                    Standard(tx),
+                )))
+            }
+            _ => todo!(),
```
jacek-kurkowski-shielded: 😄

ledger/src/test_utilities.rs
```diff
@@ -934,7 +935,7 @@ pub fn well_formed_tx_builder<
     const REWARDS_AMOUNT: u128 = 5000000000;
     let token = ShieldedTokenType(Default::default());
     let coin = CoinInfo::new(&mut rng, REWARDS_AMOUNT, token);
-    let out = ZswapOutput::new(&mut rng, &coin, 0, &secret_key.coin_public_key(), None).unwrap();
+    let out = ZswapOutput::new(&mut rng, &coin, None, &secret_key.coin_public_key(), None).unwrap();
```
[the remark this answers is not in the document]
tkerber: `None` is intended to be the weaker statement – meaning 'I don't know where this will go yet', so it's the safer upgrade choice – although it doesn't really matter either way.

ledger-wasm/src/tx.rs
tkerber: Yeah, thanks for pointing that out :face_with_peeking_eye: 
Pushed the intended error handling here.

AlexShielded (approved):
LGTM!

PR #125, opened 2025-12-18
Correct check from point at infinity to identity point.

tkerber:
NOTE: Schedule for v7, as this is technically a breaking change.

Correct the filtering of the identity element in the encryption module to correctly filter the identity, rather than assuming that the identity is the point at infinity. JubJub lacks a point at infinity, but has a finite identity element.

The point at infinity predicate is still needed as at some points we explicitly encode the point at infinity to `(0, 0)`; this is currently dead code, but is better than an assertion that the point at infinity _cannot_ occur.


AlexShielded (changes requested):
LGTM, could we get a note in the `CHANGELOG`

PR #127, opened 2025-12-18
Switch to the Midnight SRS & new midnight-zk 1.0

tkerber:


Cargo.toml
```diff
@@ -46,6 +46,18 @@ exclude = [
     "scripts/hardfork_version_patch"
 ]
 
+[patch."https://github.com/midnightntwrk/midnight-zk"]
```
AlexShielded: Could we get this switched to a permanent change, rather than a patch?

ledger/src/structure.rs
```diff
@@ -219,30 +219,121 @@ impl<D: DB> PedersenDowngradeable<D> for Pedersen {
...
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Storable, Introspection)]
 #[storable(base)]
 #[tag = "proof-versioned"]
 #[non_exhaustive]
 pub enum ProofVersioned {
-    V1(Proof),
+    V2(Proof),
 }
+
+impl Serializable for ProofVersioned {
```
AlexShielded: Would be cool to keep the derive macro and have a parameter that can get passed in to save on the boilerplate

AlexShielded (approved):
LGTM, a couple of suggestions that can be pushed until a later date. Needs a changelog update :)

Cargo.toml
tkerber: The patches are a requirement for midnight-zk right now, their team is looking at resolving this, but for now it shouldn't block us making something usable.

tkerber: Turns out they are now no longer needed.

zkir/src/ir_vm.rs
```diff
@@ -821,7 +821,7 @@ impl Relation for IrSource {
             sha3_256: false,
             keccak_256: false,
             blake2b: false,
-            nr_pow2range_cols: 4, // TODO: Get this programmatically, not as a magic constant from Miguel :)
+            nr_pow2range_cols: 1, // TODO: Get this programmatically, not as a magic constant from Miguel :)
```
miguel-ambrona: When are we gonna address this?
Shall I take care of it in a subsequent PR?

tkerber: Yeah, I think Inigo mentioned changing the handling for zkirv3, which seems reasonable to me.

PR #133, opened 2025-12-23
Add ability to compute k value of a circuit in WASM bindings of ZKIR

kapke:
Including tests against protocol circuits for the zkir v2;

### DIFF
--- a/integration-tests/src/test-objects.ts
+++ b/integration-tests/src/test-objects.ts
@@ -347,7 +347,7 @@ export const keyMaterialProvider = new (class {
 
   // eslint-disable-next-line class-methods-use-this
   async getParams(k: number): Promise<Uint8Array> {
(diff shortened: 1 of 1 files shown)


integration-tests/src/test-objects.ts
```diff
@@ -301,6 +304,53 @@ export class Static {
...
+      if (maybeIr === undefined) {
+        return undefined;
+      }
+      return maybeIr.toString('utf-8');
+    });
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  async getParams(k: number): Promise<Uint8Array> {
+    return fs.readFile(`${process.env.MIDNIGHT_PP}/bls_filecoin_2p${k}`);
```
tkerber:
```suggestion
    return fs.readFile(`${process.env.MIDNIGHT_PP}/bls_midnight_2p${k}`);
```

changed with 7.0.

tkerber (changes requested):

tkerber (approved):

PR #137, opened 2026-01-05
[PM-21051] [PM-21059] [PM-21060] Address various comments on the new `addCalls` endpoint and its interaction with `fromParts`

tkerber:
Defaults to segment 1 for zswap segments in `fromParts`, and changes error messages to be more appropriate for `addCalls`.

### DIFF
--- a/integration-tests/src/proof-worker.ts
+++ b/integration-tests/src/proof-worker.ts
@@ -11,44 +11,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { provingProvider, type ProvingKeyMaterial } from '@midnight-ntwrk/zkir-v2';
-import { readFile } from 'fs/promises';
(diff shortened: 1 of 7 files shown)


ledger-wasm/src/tx.rs
```diff
@@ -997,13 +979,15 @@ impl Transaction {
             | UnprovenWithSignatureErasedBinding(_)
             | ProvenWithSignatureBinding(_)
             | ProvenWithSignatureErasedBinding(_) => {
-                Err(JsError::new("Transaction is already bound."))?
+                Err(JsError::new("Cannot add calls to bound transaction."))?
```
jacek-kurkowski-shielded: Ohh, shouldn't it be:
`Cannot add calls to proved transaction.`?

tkerber: Nah, these cases are all `binding` cases, and that does _explicitly_ mean we can't modify them.

PR #138, opened 2026-01-05
Some fixes to benchmark runner.

tkerber:
- Something with the docker vendoring wasn't working out. Fortunately, that was a band-aid for the source code being in a private repo, so I just stripped it out.
- The `log` opcode benchmarking was only getting one data point for `map` data types, because all others were too large. I've reduced the size of the data passed in here.
- Changed a lingering filecoin reference to the midnight srs.


AlexShielded (dismissed):
LGTM

PR #147, opened 2026-01-14
fix: [PM-21238] fix benchmark pipeline

jacek-kurkowski-shielded:
We forgot to update it correctly during the repository migration, fixing it now.


tkerber:
This is likely out of date. If not, please reopen.

PR #148, opened 2026-01-14
Add zkir-v3 binary

JosephDenman:
PR https://github.com/midnightntwrk/midnight-ledger/pull/91 was merged without adding the `zkir-v3` binary to Nix flake. So, consumes like `compact-export` cannot use `zkir-v3` to generate prover and verifier keys. This PR adds the `zkir-v3` binary.


jacek-kurkowski-shielded (approved):
Approved, but we (and by “we” I mean the Ledger team) need to figure out whether we want to keep repeating the code for `zkir` or create common modules for the parts that are repeated.

cc: @tkerber @AlexShielded @jsidorenko

PR #153, opened 2026-01-23
Support for JubjubPoint in ZKIRv3

miguel-ambrona:
As described here:
https://shielded.atlassian.net/browse/PM-21413

Review commit by commit.

What is the plan w.r.t. inputs to the circuit?
It would be great if [these inputs](https://github.com/midnightntwrk/midnight-ledger/blob/ambrona@zkirv3-jubjubpoint/transient-crypto/src/proofs.rs#L706-L707) were typed too. I wonder whether I should make this change now or it requires changes in Compact and the typescript back-end.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2801,6 +2801,7 @@ dependencies = [
  "midnight-onchain-runtime",
  "midnight-serialize",
  "midnight-storage",
+ "midnight-storage-core",
  "midnight-transient-crypto",
  "midnight-zswap",
  "pastey",
@@ -2812,6 +2813,7 @@ dependencies = [
  "serde",
  "serde_json",
  "sha2",
+ "tempfile",
  "tokio",
  "tracing",
  "tracing-subscriber",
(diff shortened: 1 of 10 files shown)


iquerejeta (dismissed):
LGTM 👍  minor comments. Great to see a polymorphic add.

zkir-v3/src/ir.rs
```diff
@@ -266,6 +280,43 @@ tag_enforcement_test!(Operand);
...
+    ///
+    /// This operation will result in an error if the number of outputs
+    /// is not the exact number of raw Fr elements required to represent a
+    /// value of the input type:
+    ///
+    ///  - Native:      1 output
+    ///  - JubjubPoint: 2 outputs (x and y coordinates)
+    Encode {
+        /// The value to encode
+        input: Operand,
```
kmillikin: I called this `val` instead of `input`.

We are not consistent (`val` vs. `inputs` elsewhere), we should adopt something consistent that we like.

(I'm changing this one to `input` just to make testing easier.)

jacek-kurkowski-shielded (approved):

PR #154, opened 2026-01-23
ZKIR optimizations for `ConstrainBits`, `CondSelect`, `ReconstituteField`

miguel-ambrona:
## Summary
- Optimize `ConstrainBits` to use `assert_lower_than_fixed` range check instead of bit decomposition, skipping the constraint entirely when bits >= field size
- Optimize `CondSelect` to use direct bit conversion + select instead of `is_zero` negation trick
- Optimize `ReconstituteField` to use `linear_combination` instead of bit decomposition and reconstitution
- Add `num-bigint` dependency to both `zkir` and `zkir-v3` crates
- Increase `nr_pow2range_cols` from 1 to 4 for the new range check approach

### DIFF
--- a/.tag-decompositions/ir-source[v2-generic]
+++ b/.tag-decompositions/ir-source[v2-generic]
@@ -1,0 +1,1 @@
+((ir-minor-version[v2],u32,bool,vec(ir-instruction[v2])))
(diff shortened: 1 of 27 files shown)


iquerejeta (dismissed):
LGTM 👍

iquerejeta (dismissed):

miguel-ambrona:
@tkerber @kmillikin 

Shall we close this one? They are optimizations over ZKIRv2.

v3 will include this, but I'm not sure v2 can be changed at this stage.

AlexShielded (dismissed):
LGTM! It would be nice to get a note in the changelog 😃

tkerber (approved):
Self-review because it's just the changelog diff to approve.

PR #155, opened 2026-01-26
[PM-21558] Fix `RcMap` non-determinism

tkerber:
Using `std::collections::HashMap` iteration introduced non-determinism into `RcMap` processing. I've not yet fully thought through why this isn't isomorphic and result in the same output, but for now, forcing an ordering on it is a sufficient fix.


AlexShielded (dismissed):
LGTM, we should really look into the root of the non-determinism, would be nice to remove these sorts!

PR #157, opened 2026-01-26
Parameter update as an example.

tkerber:
## Summary
- Add `output-params` example that serializes updated ledger parameters (cost model, transaction limits) for constructing parameter update transactions via the toolkit
- Include a complete `UPDATED_COST_MODEL` with benchmark-derived picosecond timings from the ledger-7.0.0-rc.1 benchmark run
- Add `midnight-onchain-vm` as a dev-dependency and enable `fixed-point-custom-serde` feature for test builds


tkerber:
Need to polish this up, but having the *current* parameters as a reference in-repo is a good idea.

jsidorenko (dismissed):

AlexShielded (approved):

PR #158, opened 2026-01-26
Add tests to check for non-deterministic behaviour.

tkerber:
## Summary
- Add determinism checks to `TestState` by re-applying every system and user transaction and asserting identical state hashes
- Add determinism assertions to `update_rcmap` and `gc_rcmap` property tests, verifying that repeated operations produce identical hashes and removal sets


tkerber:
Tests are still valuable, need to get this polished up.

AlexShielded (approved):

PR #160, opened 2026-01-29
Ledger `6.2.0-rc.3` -> `7.0.0-rc.2` state translation

AlexShielded:
## Summary
- Implement state translation from ledger 6.2.0-rc.3 to 7.0.0-rc.2
- Add git dependencies for the old ledger (6.2.0) and new ledger (7.0.0-alpha.1) coin-structure crates
- Introduce `midnight-circuits` v5.0.2 as an additional dependency alongside v6.0.0 for the translation layer
- Update Cargo.lock with new/renamed crates and dual-version dependencies


tkerber:
Need to revisit; we need to plan where to keep these, and this one *should* be kept as a reference, even if we won't use it.

AlexShielded:
> Need to revisit; we need to plan where to keep these, and this one _should_ be kept as a reference, even if we won't use it.

Is it worth porting it to some sort of `examples/` folder, with pinned ledger versions?

PR #164, opened 2026-02-03
Extract `storage-core` from `storage`

AlexShielded:
Some `pub(crate)`s had to be swapped for `pub`s. Similarly the suite of functions implemented as part of the MPT under `Sp<Node>` had to be swapped from `fn foo(&self,...)` to `fn foo(sp: &Sp<Node>,...)` in order to resolve an issue where the struct and the impl live in different places (it didn't make sense to move MPT `Node` to `storage-core`, nor `Sp` to `storage`).

### DIFF
--- a/storage-core/src/arena.rs
+++ b/storage-core/src/arena.rs
@@ -1954,6 +1954,7 @@ pub mod bin_tree {
     #[derive_where(Clone, PartialEq, Eq)]
     #[tag = "test-bin-tree"]
     #[storable(db = D)]
+    /// A binary tree used for stress-testing
     pub struct BinTree<D: DB> {
         value: u64,
         pub(crate) left: Option<Sp<BinTree<D>, D>>,
@@ -1972,6 +1973,7 @@ pub mod bin_tree {
     }
 
     impl<D: DB> BinTree<D> {
+        /// Create a new `BinTree`
         pub fn new(
             value: u64,
             left: Option<Sp<BinTree<D>, D>>,


storage-core/src/backend.rs
```diff
@@ -1036,7 +1036,8 @@ pub struct OnDiskObject<H: WellBehavedHasher> {
     /// `StorageBackend::get_root_count` for details -- but not here in the
     /// object!
     pub ref_count: u64,
-    pub(crate) children: std::vec::Vec<ArenaKey<H>>,
+    /// Used in `storage` crate
```
tkerber: Proper doc please. This not being `pub` has been an annoyance to Heiko for building a DB outside the crate as well.

storage-core/src/storage.rs
tkerber: Unfortunate that git didn't idenity this as a move-and-create-replacement :sweat_smile:

storage/src/lib.rs
```diff
@@ -1,57 +1,5 @@
-// This file is part of midnight-ledger.
```
tkerber: License header missing?

storage/src/merkle_patricia_trie.rs
tkerber: Nit: These still would probably be nicer as associated functions on `Node` than bare functions in the MPT module? It's also causing this to be a massive diff for largely benign reasons.

tkerber:
Clippy also seems to be throwing a fit

storage/src/merkle_patricia_trie.rs
AlexShielded: Sounds good to me

tkerber (approved):

PR #165, opened 2026-02-04
Implement Schnorr signatures over JubJub.

iquerejeta:
This PR implements the wasm bindings counterpart for the compact implementation of Schnorr signatures over Jubjub https://github.com/LFDT-Minokawa/compact/pull/65. 

While the verification is not strictly needed, I've added for completeness.


onchain-runtime-wasm/src/primitives.rs
```diff
@@ -399,3 +402,125 @@ pub fn ec_mul_generator(val: JsValue) -> Result<JsValue, JsError> {
...
+
+/// Converts a BLS12-381 scalar field element into a Jubjub scalar field element,
+/// reducing modulo the Jubjub scalar field modulus.
+fn jubjub_scalar_from_native(native: outer::Scalar) -> embedded::Scalar {
+    let mut wide = [0u8; 64];
+    wide[..32].copy_from_slice(&native.to_bytes_le());
+    embedded::Scalar::from_bytes_wide(&wide)
+}
+
+/// Converts a BLS12-381 scalar field element into a Jubjub scalar field element.
```
miguel-ambrona: Wrong doc.

miguel-ambrona (dismissed):
LGTM.

I see a common pattern in all tests that maybe we can share in a helper function:
```rust
    let mut rng = StdRng::seed_from_u64(_);
    let sk: EmbeddedFr = rng.r#gen();
    let sig = sign(&mut rng, sk, &msg);
```

transient-crypto/src/fab.rs
```diff
@@ -168,11 +168,7 @@ impl TryFrom<&ValueAtom> for EmbeddedFr {
     type Error = InvalidBuiltinDecode;
 
     fn try_from(value: &ValueAtom) -> Result<EmbeddedFr, InvalidBuiltinDecode> {
-        if value.0.len() <= FR_BYTES {
-            EmbeddedFr::from_le_bytes(&value.0).ok_or(InvalidBuiltinDecode("EmbeddedFr"))
-        } else {
-            Err(InvalidBuiltinDecode("EmbeddedFr"))
-        }
+        EmbeddedFr::from_le_bytes_wide(&value.0).ok_or(InvalidBuiltinDecode("EmbeddedFr"))
```
JosephDenman, line 171: The old code used `from_le_bytes` which rejects values `>=` the embedded field modulus. The new code uses `from_le_bytes_wide` which silently reduces them. This is a potentially breaking behavioral change that affects all callers of `EmbeddedFr::try_from(value_atom)`, not just the Schnorr code. Previously invalid inputs will now succeed with a different value. Was this intentional? If so, it deserves a comment explaining why and an audit of downstream callers.

transient-crypto/src/schnorr.rs
```diff
@@ -0,0 +1,183 @@
...
+        announcement,
+        response,
+    }
+}
+
+/// Verifies a Schnorr signature over the Jubjub curve.
+pub fn verify(pk: EmbeddedGroupAffine, msg: &[Fr], sig: &SchnorrSignature) -> bool {
+    // Unwraps cannot fail as the inner curve is an edwards curve, and the
+    // identity is part of the curve (i.e., identity also has coordinates).
+    let pk_x = pk.x().unwrap();
```
JosephDenman, line 80: `x` is defined:

```
    /// Retrieves the curve point's affine `x` coordinate.
    /// Or `None` if this is the identity
    pub fn x(&self) -> Option<Fr> {
        Into::<embedded::AffineExtended>::into(self.0)
            .coordinates()
            .map(|c| Fr(c.0))
    }
```

So, apparently it returns `None` if it is the identity. Even if `None` isn't returned from `x` today on identity, if the `coordinates()` implementation changes to match its docs, this will panic. Maybe handle the `None` case explicitly?

onchain-runtime-wasm/src/primitives.rs
```diff
@@ -399,3 +402,35 @@ pub fn ec_mul_generator(val: JsValue) -> Result<JsValue, JsError> {
...
+    let fr = curve::Fr::try_from(&*val)?;
+    let mut wide = [0u8; 64];
+    wide[..32].copy_from_slice(&fr.0.to_bytes_le());
+    let embedded = embedded::Scalar::from_bytes_wide(&wide);
+    Ok(to_value(&Value::from(EmbeddedFr(embedded)))?)
+}
+
+#[wasm_bindgen(js_name = "nativeFromJubjubScalar")]
+/// Converts a JubJub scalar field element to a native field element (BLS12-381 scalar).
+pub fn native_from_jubjub_scalar(jubjub: JsValue) -> Result<JsValue, JsError> {
```
JosephDenman, line 443: Looks like this wasn't added to `onchain-runtime-v3.d.ts`.

transient-crypto/src/schnorr.rs
```diff
@@ -0,0 +1,183 @@
...
+
+    let challenge = compute_challenge(ann_x, ann_y, pk_x, pk_y, msg);
+
+    let lhs = EmbeddedGroupAffine::generator() * sig.response;
+    let rhs = sig.announcement + pk * challenge;
+
+    lhs == rhs
+}
+
+/// Computes the verfying key from a secret key.
```
JosephDenman:
```suggestion
/// Computes the verifying key from a secret key.
```

JosephDenman (changes requested):
Mostly documentation requests. But the onchain runtime WASM API is actually incomplete.

transient-crypto/src/fab.rs
iquerejeta: Yes, this is intentional. Any value that previously succeeded from_le_bytes (i.e. was already canonical, < field modulus) will produce the identical result with from_le_bytes_wide, so there is no breaking change for existing callers. The switch to wide reduction is in fact the correct behavior: without it, there is a bug in the current codebase that prevents users from producing valid Schnorr signatures. Additionally, `TryFrom<&ValueAtom> for EmbeddedFr` is only ever used in the context of JubJub scalars, and JubJub scalars are only meaningful as inputs to JubJub point multiplication — matching the behavior of the Compact implementation.

JosephDenman (dismissed):
LGTM

onchain-runtime-wasm/onchain-runtime-v3.d.ts
```diff
@@ -598,6 +598,22 @@ export function proofDataIntoSerializedPreimage(
  */
 export function bigIntModFr(x: bigint): bigint;
```
tkerber, line 600: Suggest also adding `export function maxJubjubScalar(): bigint` and `export function bigIntModJubjubScalar(x: bigint): bigint`

tkerber (approved):

PR #190, opened 2026-02-16
Optimize Sp allocations to minimize cache & disk interaction

tkerber:
Also boosts how far the night benchmark can be pushed by trimming everything that forces a full in-memory load during init.


AlexShielded (approved):
LGTM

PR #191, opened 2026-02-17
Fix Zswap past tree roots retention period

kapke:
The `post_block_update` function had hardocded 1h as the retention period, while it should use the global TTL limit value. This PR fixes that.


tkerber (dismissed):

jsidorenko (dismissed):

jsidorenko (commented):
Nice!

tkerber (dismissed):
Great! I think we now can actually merge this with the new branch structure. A changelog entry would be nice, but I'm happy to merge regardless.

kapke:
Removed the changes to maintain backward compatibility and added a changelog entry.

tkerber (approved):

AlexShielded (approved):
LGTM

PR #201, opened 2026-02-19
Prepare ledger `8.0.0-rc.4`

AlexShielded:


CHANGELOG_storage.md
```diff
@@ -1,5 +1,9 @@
 # `storage` Changelog
 
+## Version `2.0.1`
+
+- feat: expose optimisations in `storage-core` `1.0.2`
```
tkerber: Technically that doesn't need its own release. But we *did* change the `rcmap` internals to use `BTreeMap` instead of `HashMap`, which does.

PR #205, opened 2026-02-26
fix: upgrade rollup to fix CVE-2026-27606

gilescope:


tkerber:
Nothing left after merge, I got to it via renovate first :grimacing:

PR #220, opened 2026-03-03
Generalize `ecMul` in transient-crypto to support scalars over the Jubjub order

miguel-ambrona:
Fixes https://github.com/midnightntwrk/midnight-security/issues/64

ZKIRv2 does not have types and thus the signature of function `ecMul` expects a "native" scalar (from BLS12-381) instead of a Jubjub scalar.

The internal functions for parsing native scalars as Jubjub scalars failed (before this PR) if the input was bigger than the Jubjub order. Some important use cases (such as verifying Schnorr signatures over Jubjub) require this cast to not fail.

This is not a breaking change for the ledger.

### DIFF
--- a/CHANGELOG_transient-crypto.md
+++ b/CHANGELOG_transient-crypto.md
@@ -1,5 +1,9 @@
 # `transient-crypto` Changelog
 
+## Version `2.0.1`
+
+- bug fix: generalize ecMul to support scalars over the Jubjub order
+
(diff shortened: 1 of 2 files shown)


AlexShielded (dismissed):
Looks good to me. Could you add a note to the changelog?

iquerejeta:
Fixed by #165

PR #230, opened 2026-03-18
[PM-22358] fix: serialize_to_node_list panic on mixed Direct/Ref children

mpskowron:
serialize_to_node_list_bounded used ArenaKey as HashMap keys, causing
  nodes with the same content hash but different ArenaKey variants
  (Direct vs Ref) to be treated as distinct entries. This led to a
  "node must be present" panic when both variants appeared in the same
  DAG.

  Also fix into_tracked to increment the metadata ref count for the
  returned Ref Sp. - not sure if this is correct

### DIFF
--- a/CHANGELOG_storage_core.md
+++ b/CHANGELOG_storage_core.md
@@ -1,5 +1,10 @@
 # `storage-core` Changelog
 
+## Unreleased
+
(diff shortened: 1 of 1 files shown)


tkerber (dismissed):
I think `into_tracked` might indeed still be off. I'll take a look later, it should probably mirror / be factored out of `persist`'s promotion logic.

PR #234, opened 2026-03-23
Incremental garbage collection

tkerber:
Adds `gc-v1` feature to `storage-core`, which enables an incremental garbage collector.


AlexShielded (changes requested):
Looks good! Would be neat to have some simple GC tests that show the iterative GC working on a toy example

AlexShielded (approved):
LGTM!

PR #250, opened 2026-03-27
feat: parity-db log pipeline processing trigger

Klapeyron:
## Summary
- Add `run_log_pipeline()` method to the `DB` trait and `StorageBackend` to drive ParityDB's internal log processing pipeline on demand
- Implement `run_log_pipeline()` for the ParityDB backend, calling `process_commits()`, `flush_logs()`, and `enact_logs()` in sequence
- Enable the `parity-db/instrumentation` feature flag for the parity-db feature
- Thread `run_log_pipeline()` through `WrappedDB` delegation

### DIFF
--- a/storage-core/Cargo.toml
+++ b/storage-core/Cargo.toml
@@ -10,7 +10,7 @@ default = []
 
 proptest = ["dep:proptest", "dep:proptest-derive", "base-crypto/proptest"]
 sqlite = ["dep:rusqlite", "dep:r2d2_sqlite", "dep:r2d2", "dep:fs2"]
-parity-db = ["dep:parity-db"]
+parity-db = ["dep:parity-db", "parity-db/instrumentation"]
 bench = ["dep:criterion", "dep:pprof"]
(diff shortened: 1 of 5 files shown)


gilescope (dismissed):
This works. I would think alternatively it might well be possible to turn this feature on just in the node repo where it's used. Being explicit is less surprising though.

tkerber (commented):
`run_log_pipeline` is a bit of an opaque name... I wish we had a better one, but I won't block this on it. You do need to mark this as not a draft, however!

Klapeyron:
yea I renamed it to `flush_log_pipeline`, not great not terrible I hope but sounds closer to what it in reality does. Sorry for the draft, wanted to make it for review once we reproduce and check in the env (there is unexpected delay with that)

tkerber:
Need clarity if this is still desired, after additional work demonstrated that the substrate DB is the culprit of practical issues.

Klapeyron:
yes, we can close it

PR #252, opened 2026-03-27
Implement ECDSA in base-crypto

iquerejeta:
Implemented ECDSA to be added in the future to the ledger. The ECDSA implementation follows the style and structure of Schnorr. 

We've renamed the `signatures` module to `schnorr`.


tkerber (dismissed):
Note that this is semver breaking (specifically of `base_crypto::signature`). Breaking changes to `base-crypto` are really expensive in practice because it's a dependency of `storage-core`, which we want to unify across versions.

tkerber:
Part of #412

tkerber (approved):

PR #256, opened 2026-03-27
Formally switch version to 9.

tkerber:
## Description

Bumps version on the ledger 9 branch to ledger-9. Switches to epoch versioning, so ledger-9 inits at version 0.1.0.

## Sanity Checklist

This PR:


AlexShielded (changes requested):
The release workflow will need change to reflect the versioning change. ATM if we tried to release a version off this branch it would release 0.1.0

AlexShielded (dismissed):
Nice!

AlexShielded (approved):

PR #279, opened 2026-03-31
More flexible `findPathForLeaf`

tkerber:
## Description

`findPathForLeaf` only worked on values inserted into Merkle trees with `insert`, and not `insertHash`. While we relied on `pathForLeaf` to be used for other Merkle trees, and to be generally preferable due to it not performing an O(n) scan of the tree, we didn't provide good mechanisms for contracts to determine which index elements actually ended up with.

The solution is to make `findPathForLeaf` more expressive. This PR allows it to scan ranges in the tree, and to search for hashes rather than elements.

## Sanity Checklist

This PR:

### DIFF
--- a/CHANGELOG_onchain-runtime.md
+++ b/CHANGELOG_onchain-runtime.md
@@ -1,5 +1,10 @@
 # `midnight-onchain-runtime` Changelog
 
+## Unreleased
+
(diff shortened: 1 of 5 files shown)


AlexShielded (approved):
Nice!

Issue #276, opened 2026-03-31
Enable shielded wallet sync via `applyTransaction` and collapsed Merkle tree updates


kapke:
## Describe the proposed feature:

Expose API, which allows shielded wallet to sync via applying raw transactions or applying raw transactions interleaved with collapsed Merkle tree updates. 

## User Need:

Such API enables a significantly more efficient synchronisation at a cost of encryption secret key exposure to the indexer. Indirectly, it also allows to fast-track synchronisation of newly created wallets.

## Details:

Following methods are needed in the WASM bindings. Together they allow shielded wallet to evolve its state via applying transactions/offers and stop tracking coins known to be spent:
- `ZswapLocalState.applyTransactionWithChanges(secretKeys: ZswapSecretKeys, tx: Transaction, result: TransactionResult): [ZswapLocalState, ZswapChanges]`
- `ZswapLocalState.applyWithChanges(secretKeys: ZswapSecretKeys, tx: Offer, transaction hash: Hash): [ZswapLocalState, ZswapChanges]`
- `ZswapLocalState.merkleTreeRoot(): Hash`
- `ZswapLocalState.removeCoinByNullifier(nullifier: Nullififer): ZswapLocalState`

PR #280, opened 2026-03-31
Extend wallet-level control and visibility into events

tkerber:
## Description

Adds a swathe of endpoints for the wallet to more granularly modify its internal state, as well as the ability to inspect Zswap/Dust events directly.

## Sanity Checklist

This PR:


kapke:
Looking at #276 I can see that 2 things are missing:
1. `withChanges` variants of the `apply`/`applyTransaction` methods to keep the transaction history capabilities.
2. a getter for the Merkle tree root so that wallet has a cheap way to verify data integrity

jsidorenko (dismissed):

jacek-kurkowski-shielded (approved):
LGTM!

PR #309, opened 2026-04-03
feat: add LedgerState::apply_guaranteed_only function, which lets the user apply only guaranteed part of the transaction

mpskowron:
## Description

Add apply_guaranteed_only to LedgerState — applies only the guaranteed segment (segment 0) of a transaction, skipping all fallible segments. apply is refactored to delegate segment 0 processing to apply_guaranteed_only.
This will be used by midnight-node for pre-dispatch/mempool validation, where the accept/reject decision depends solely on the guaranteed segment. If we pair this change with tuning of (time_to_disimss_per_byte x max_tx_size) params we should be able to cap midnight-node tx verification time. If we cap it low enough, it should fix the https://github.com/midnightntwrk/midnight-security/issues/29 . 

## Sanity Checklist

This PR:


jsidorenko (commented):
Nice!

ozgb (dismissed):

ledger/src/semantics.rs
```diff
@@ -1208,26 +1208,27 @@ impl<D: DB> LedgerState<D> {
...
                 let cloned_stx = stx.clone();
                 let segments = cloned_stx.segments();
                 let mut segment_success = std::collections::BTreeMap::new();
-                let mut events = Vec::new();
+                segment_success.insert(0, Ok(()));
                 let mut total_success = true;
-                let mut new_st = self.clone();
-                for &segment in segments.iter() {
+
+                for &segment in segments.iter().filter(|&&s| s != 0) {
```
Klapeyron: maybe it is worth to introduce helper to `StandardTransaction`
```rust
    fn segment_ids_iter(&self) -> impl Iterator<Item = u16> + use<'_, S, P, B, D> {
        self.intents.keys().chain(self.fallible_coins.keys())
    }

    pub fn segments(&self) -> Vec<u16> {
        let mut segments = once(0).chain(self.segment_ids_iter()).collect::<Vec<_>>();
        segments.sort();
        segments.dedup();
        segments
    }

    pub fn fallible_segments(&self) -> Vec<u16> {
        let mut segments = self
            .segment_ids_iter()
            .filter(|&s| s != 0)
            .collect::<Vec<_>>();
        segments.sort();
        segments.dedup();
        segments
    }
```

ledger/src/semantics.rs
```diff
@@ -1208,26 +1208,27 @@ impl<D: DB> LedgerState<D> {
     ) -> (Self, TransactionResult<D>) {
         let res = match &tx.inner {
             Transaction::Standard(stx) => {
+                let (mut new_st, result) = self.apply_guaranteed_only(tx, context);
+                let TransactionResult::Success(mut events) = result else {
+                    return (new_st, result);
+                };
+
                 let cloned_stx = stx.clone();
                 let segments = cloned_stx.segments();
```
Klapeyron:
```suggestion
                let segments = stx.segments();
```

mpskowron: changed

ledger/src/semantics.rs
mpskowron: applied partially. Added fallible_segments method. I didn't extract "segemnt_ids_iter" as I couldn't come up with a good name for it. "segment_ids_iter" might be misleading to the reader given that semgents and fallible_segments methods return sorted and deduplicated semgement ids.

Klapeyron: we can try to change iter -> map pattern to keys() function call

mpskowron: done

ledger/src/semantics.rs
```diff
@@ -1206,42 +1291,51 @@ impl<D: DB> LedgerState<D> {
...
-            }
-            Transaction::ClaimRewards(rewards) => claim_unshielded::<D>(
+    #[instrument(skip(self, tx, context), fields(tx = ?tx.hash))]
+    pub fn apply_guaranteed_only(
+        &self,
+        tx: VerifiedTransaction<D>,
+        context: &TransactionContext<D>,
+    ) -> Result<GuaranteedApplyResult<D>, TransactionInvalid<D>> {
+        let is_standard = matches!(tx.inner, Transaction::Standard(_));
+        if is_standard {
```
Klapeyron: here we make matches call followed by the if - else statement, shall we make just a match over tx.inner?

tkerber: Oh yeah, I missed Claude writing that horror.

ledger/src/semantics.rs
```diff
@@ -1269,6 +1309,62 @@ impl<D: DB> LedgerState<D> {
...
+            }
+            Transaction::ClaimRewards(rewards) => {
+                let (state, result) = claim_unshielded::<D>(
+                    self,
+                    rewards,
+                    &context.block_context,
+                    EventSource {
+                        transaction_hash: tx.hash,
+                        logical_segment: 0,
+                        physical_segment: 0,
```
Klapeyron: one more possible
```suggestion
                        logical_segment: GUARANTEED_SEGMENT,
                        physical_segment: GUARANTEED_SEGMENT,
```

AlexShielded (approved):

PR #310, opened 2026-04-06
Rehashing serde deserialized merkle trees

AlexShielded:
## Description

Rehashing serde deserialized merkle trees, to address https://github.com/midnightntwrk/midnight-security/issues/46

## Sanity Checklist

This PR:
- [N] contains changes to transaction behaviour [^1]
- [N] contains changes to architecture [^1]
- [N] contains breaking API changes [^1][^2]
- [N] contains data format changes [^1]
- [N] contains circuit behaviour changes [^3]
- [??] requires a backport to prior versions
- [N] is primarily authored by AI
- [Y] I have self-reviewed the PR diff
- [Y] changelog and package versions have been amended, or don't require it


tkerber (dismissed):

transient-crypto/src/merkle_tree.rs
```diff
@@ -459,6 +459,7 @@ impl<'de, A: Deserialize<'de> + Storable<D>, D: DB> Deserialize<'de> for MerkleT
         data.into_iter()
             .try_fold(MerkleTree::blank(height), |mt, (k, (v, a))| {
                 MerkleTree::update_hash(&mt, k, v, a)
+                    .map(|mt| mt.rehash())
```
jsidorenko: maybe it's better to re-hash the tree in the end instead of on each iteration?

AlexShielded: Right you are!

tkerber (approved):

PR #392, opened 2026-04-13
Copy node team's homework on renovate.

tkerber:
## Description

Replace our old renovate config with one mostly copied from the node repo. Hopefully will be more sane than what just rand this morning...

## Sanity Checklist

This PR:


AlexShielded (approved):
Looks sensible

PR #407, opened 2026-04-14
chore: included additional application cost

Klapeyron:
## Sanity Checklist

This PR:


tkerber (changes requested):
Not mergeable against `ledger-8`.

tkerber:
Capturing the spirit of this under #414 – we *do* need to adjust the cost model, but this *will* require a hard fork, and we likely want to change what we actually do prior to that. I don't think this will be the simple PR you're envisioning here.

PR #422, opened 2026-04-14
fix: include bridge_receiving in NIGHT balance invariant

### LINKED ISSUE #411: Bridge is not subject to NIGHT invariant
### Context & versions:
The (currently unused) token bridge system transactions cannot satisfy the NIGHT invariant, as the `bridge_receiving` map does not carry invariant annotations. As a result, any system transaction modifying it will be rejected.

### Steps to reproduce:

see above

### Actual behavior:

the bridge transaction is rejected

### Expected behavior:

the bridge transaction is accepted

cosmir17:
## Summary
`bridge_receiving` Map was excluded from `check_night_balance_invariant`, causing `DistributeNight(CardanoBridge)` system transactions to fail with an invariant violation. This blocks all Cardano-to-Midnight bridge transfers at or above the minimum threshold.

## Notes
- Existing bridge claim tests (`simple_bridge_claim`, `bridge_claim_not_enough_available`) continue to pass
- Thomas milestoned #411 to ledger-9.0.0 but the fix is identical on both branches. Targeting ledger-8 to potentially unblock bridge work sooner. Happy to retarget to ledger-9 if preferred.
- The serialization tag change (v13 to v14) means existing serialized state with `bridge_receiving` using `SizeAnn` would need migration. Since `bridge_receiving` is empty on all live networks (bridge has never been active), migration should be straightforward.

Fixes: #411

### DIFF
--- a/.tag-decompositions/ledger-state[v14]
+++ b/.tag-decompositions/ledger-state[v14]
@@ -1,0 +1,1 @@
+((string,#ledger-parameters[v5],u128,mpt-map(user-address[v1],u128,night-annotation),u128,u128,mpt-map(user-address[v1],u128,night-annotation),mpt-map(token-type[v1],u128,size-annotation),#zswap-ledger-state[v5],mpt-map(contract-address[v2],contract-state[v6],night-annotation),#unshielded-utxo-state[v2],replay-protection-state[v1],#dust-state[v1]))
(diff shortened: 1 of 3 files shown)


tkerber:
> Thomas milestoned https://github.com/midnightntwrk/midnight-ledger/issues/411 to ledger-9.0.0 but the fix is identical on both branches. Targeting ledger-8 to potentially unblock bridge work sooner. Happy to retarget to ledger-9 if preferred.

It's not possible to do thing without it being a hard fork, as it changes the state. That's why it needs to be targeted for ledger 9.

cosmir17:
Thanks @tkerber, that makes sense. The NightAnn addition changes the serialization format (v13 → v14), so it can't go into ledger-8 without a hard fork. Will retarget to ledger-9. I close this PR. This branch is not for ledger 9.

PR #429, opened 2026-04-16
refactor: introduced type alias for segments

Klapeyron:
## Description

Introduced type alias for segment and guaranteed segment associated constant

## Sanity Checklist

This PR:

### DIFF
--- a/ledger/src/structure.rs
+++ b/ledger/src/structure.rs
@@ -1555,13 +1555,8 @@ impl<S: SignatureKind<D>, P: ProofKind<D>, B: Storable<D>, D: DB> Intent<S, P, B
 /// Logical segment index used in standard transactions and ledger application order.
 pub type Segment = u16;
 
-/// Namespace for [`Segment`] constants.
-pub struct Segments;
-
-impl Segments {
-    /// Segment id for the guaranteed phase (fees, guaranteed coins, guaranteed transcripts).
-    pub const GUARANTEED: Segment = 0;
-}
+/// Segment id for the guaranteed phase (fees, guaranteed coins, guaranteed transcripts).
+pub const GUARANTEED_SEGMENT: Segment = 0;
 
 #[derive(Storable)]
 #[storable(db = D)]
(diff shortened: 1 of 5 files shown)


spec/intents-transactions.md
```diff
@@ -724,7 +724,7 @@ impl LedgerState {
                     self = state;
                     segment_success = segment_success.insert(segment, true);
                 }
-                Err(e) => if segment == 0 {
+                Err(e) => if segment == Segments::GUARANTEED {
```
Klapeyron: there exists alternative syntax with matches, not sure which one will be better

```suggestion
                Err(e) => if matches!(segment, Segments::GUARANTEED) {
```

ledger/src/structure.rs
```diff
@@ -1552,23 +1552,34 @@ impl<S: SignatureKind<D>, P: ProofKind<D>, B: Storable<D>, D: DB> Intent<S, P, B
...
 }
 
+/// Logical segment index used in standard transactions and ledger application order.
+pub type Segment = u16;
+
+/// Namespace for [`Segment`] constants.
+pub struct Segments;
+
+impl Segments {
+    /// Segment id for the guaranteed phase (fees, guaranteed coins, guaranteed transcripts).
```
Klapeyron: please help me find appropriate description

tkerber:
Please motivate this change.

Klapeyron:
It is trying to improve visibility of segment 0 threaten as guaranteed, I think representing it as associated constant is a good match due we have flexibility of underlying u16 and well defined special value

ledger/src/structure.rs
```diff
@@ -1552,23 +1552,34 @@ impl<S: SignatureKind<D>, P: ProofKind<D>, B: Storable<D>, D: DB> Intent<S, P, B
     }
 }
 
+/// Logical segment index used in standard transactions and ledger application order.
+pub type Segment = u16;
+
+/// Namespace for [`Segment`] constants.
+pub struct Segments;
```
tkerber: I do not like this for several reasons:
- It would add both `Segment` and `Segments` types, which is confusing
- `Segments` being used as a namespace seems silly, rust has a primitive for namespacing, and that's the module.
- `Segments` is a misleading name when it contains aliases for _exactly one_ segment.
- This being namespaced is inconsistent with most other constants and markers used in the codebase.

tkerber (changes requested):

ledger/src/structure.rs
Klapeyron: I changed it to standalone constant and type alias https://github.com/midnightntwrk/midnight-ledger/pull/429/commits/2f909870c335ed601e70302a15184eee971e8f1d

tkerber (approved):

PR #430, opened 2026-04-16
feat: allow shared parity_db backend through generic Deref

ozgb:
## Description

Allows for using a shared reference of ParityDb via smart pointers, e.g. RwLock.

Fixes a bug in insert_node where `COLUMN_OFFSET` was not respected.

## Sanity Checklist

This PR:


storage-core/src/db/sql.rs
ozgb, line 1: Changed due to `cargo fmt --all` - can revert

tkerber (approved):

Issue #432, opened 2026-04-16
StorageBackend::get() panics via cache_insert_new_key when pending Update delta exists in memory


cosmir17:
## Symptom
chain-indexer panics with:

  panicked at midnight-storage-core/src/backend.rs:1161:9:
  key must not already be in memory

Hit on a plain V0_22 block on qanet (block 11074) with indexer 4.1.0-0c4830f4 running against node 0.22.5-rc.1. Symptom class matches midnightntwrk/midnight-indexer#987 but the trigger is different (not a v1.0 runtime upgrade).

Pod crash-looped on restart: backward-traverse, re-index to block 11074, panic again.

## Root cause

`StorageBackend::get()` cache-miss branch (backend.rs:381-398) reads the object from DB, then checks `peek_from_memory` for an existing `CacheValue::Update { delta }` and builds a `ReadAndUpdate` variant without removing the existing `Update` entry. The subsequent `cache_insert_new_key` then trips its `"key must not already be in memory"` precondition.

Same class of bug as 7ee86557 fixed in `StorageBackend::cache()`, but `get()` was not patched.

Visible symptom in dev-profile builds where `debug_assert!` is live. In release builds, `debug_assert!` is stripped and the next `assert!` (`"write cache is unbounded, it can't evict"`) would fire instead when the new pending value tries to replace the existing one in the write cache.

## Reproduction

Regression test in PR #433 (`get_with_pending_update_in_memory`, InMemoryDB + SqlDB variants) reproduces the panic at the same file/line/message.

Also reproduced end-to-end against qanet chain by running chain-indexer locally with a `[patch.crates-io]` pointing at the fix branch — indexer now indexes past block 11074 without panic.

## Fix

midnightntwrk/midnight-ledger#433 (merged): mirror 7ee86557 — remove the pending `Update` from memory before constructing `ReadAndUpdate` and calling `cache_insert_new_key`.

## Target release (? feel free to update the ticket 🙏)

Ledger 8.1 (per 17 Apr release call). Thomas to backport #433 to `ledger-8.1` and cut `ledger-8.1.0-rc.2` / `storage-core-1.2.0-rc.2`. Node and indexer then reintegrate.

The concrete trigger was observed from the indexer side; whether the node's normal workload also reaches the same sequence under layout-v2 is less certain. Node team may want to confirm independently, though reintegrating the fixed dep looks like the safer default either way.

## Related

[...]

PR #433, opened 2026-04-16
fix: remove pending Update from memory before cache_insert_new_key in get()

cosmir17:
## Summary

Candidate fix for the `cache_insert_new_key` `"key must not already be in memory"` panic, same class as 7ee86557 fixed in `cache()`, but in `get()`.

Related: midnight-ledger #432, midnight-indexer #987 (reoccurring on qanet-blue with indexer 4.1.0-0c4830f4 at block 11074, plain V0_22). Happy to be wrong on root cause, Alex is the expert here.

## What was happening

On the `get()` cache-miss branch, when a pending `CacheValue::Update { delta }` was already in memory for the requested key, the function built `ReadAndUpdate` and called `cache_insert_new_key` without removing the existing `Update` first. Dev builds trip the `debug_assert!`. Release builds would hit the `assert!` `"write cache is unbounded, it can't evict"` immediately after, so not dev-only.

## How it's caused

`update_counts` can put a pure `Update { delta }` into memory for a key not previously cached. A later `get()` on that key hits the precondition violation.

## The fix

Mirror 7ee86557: remove the pending `Update` before constructing `ReadAndUpdate`. 3-line change.

## Testing

Added `get_with_pending_update_in_memory` regression test (InMemoryDB + SqlDB). Without the fix it panics at `backend.rs:1161:9` with the exact midnight-indexer #987 symptom. With the fix, 53/53 tests pass.

## Test plan

- [x] `cargo test -p midnight-storage-core --features "layout-v2,sqlite"` — 53/53
- [ ] Review whether this is the actual qanet-blue trigger or a separate/additional path
- [ ] If accepted: cut storage-core RC, bump indexer, retest on qanet-blue


AlexShielded (approved):
LGTM!

PR #439, opened 2026-04-17
Prepare `storage-core` `1.2.0-rc.2`

AlexShielded:
## Description

Cherrypicking:

- https://github.com/midnightntwrk/midnight-ledger/pull/430
- https://github.com/midnightntwrk/midnight-ledger/pull/433

## Sanity Checklist

This PR:


storage-core/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "midnight-storage-core"
-version = "1.2.0"
+version = "1.3.0"
```
tkerber: We don't need to bump this: `1.2.0` is in `rc` (the tag is `1.2.0-rc.1`), so this would be a `1.2.0-rc.2`. The reason the `rc` doesn't show up here is to allow cargo to still resolve it.

tkerber (changes requested):

storage-core/Cargo.toml
AlexShielded: Ah, I see

tkerber (approved):

Issue #416, opened 2026-04-14
Allow recording of contract IR on-chain


tkerber:
## Describe the proposed feature:

Extend `ContractOperation` with an optional field for ZKIR. This should have include the type header for compatibility with any past and future supported IRs. Initially, this serves *no purpose* besides advise, but should encode the IR for this specific operation. This may be used for smart contracts surviving proof system upgrades in the future.

## User Need:

For smart contracts to survive proof system upgrades without trusted intervention.

## Expected Benefit:

Smart contracts are theoretically upgradable, as their circuit semantics are kept on-chain, instead of just a commitment to them.

## Details:

tbd

jacek-kurkowski-shielded:
This one has been verified & tested by: https://github.com/midnightntwrk/midnight-ledger/issues/584

PR #448, opened 2026-04-21
Allow recording of contract IR on-chain

AlexShielded:
## Description

Satisfies https://github.com/midnightntwrk/midnight-ledger/issues/416

## Sanity Checklist

This PR:


onchain-state/src/state.rs
```diff
@@ -864,16 +865,17 @@ impl<D: DB> Default for ContractState<D> {
     Serializable, Clone, Hash, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize, Storable,
 )]
 #[storable(base)]
-#[tag = "contract-operation[v4]"]
+#[tag = "contract-operation[v5]"]
 #[non_exhaustive]
 pub struct ContractOperation {
     pub v2: Option<VerifierKey>,
+    ir: Option<IrSource>,
```
tkerber: I believe this should be a generic container (i.e. `Sp<Vec<u8>>`) instead. We don't want to draw a dependency on the IR itself.

onchain-state/Cargo.toml
```diff
@@ -16,6 +16,7 @@ transient-crypto = { version = "2.0.0", path = "../transient-crypto", package =
 base-crypto = { version = "1.0.0", path = "../base-crypto", package = "midnight-base-crypto" }
 storage = { version = "2.0.0", path = "../storage", package = "midnight-storage" }
 serialize = { version = "1.0.0", path = "../serialize", package = "midnight-serialize" }
+zkir = { version = "2.1.0", path  = "../zkir", package = "midnight-zkir" }
```
tkerber: :-1:

tkerber (changes requested):

test_cost_model_fix.rs
tkerber, line 1: The hell is this garbage 😅

onchain-state/src/state.rs
```diff
@@ -864,16 +864,17 @@ impl<D: DB> Default for ContractState<D> {
     Serializable, Clone, Hash, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize, Storable,
 )]
 #[storable(base)]
-#[tag = "contract-operation[v4]"]
+#[tag = "contract-operation[v5]"]
 #[non_exhaustive]
 pub struct ContractOperation {
     pub v2: Option<VerifierKey>,
+    ir: Option<Sp<Array<u8>>>,
```
tkerber: :-1: This is possibly the least efficient way we could represent a bitstring.`Vec<u8>` with size limit is the way to go.

ledger/src/structure.rs
```diff
@@ -1879,6 +1883,20 @@ where
...
+                                // Use a conservative estimate for the verifier key size during validation
+                                // to prevent DoS attacks from oversized metadata
+                                let entry_point_size = call.entry_point.serialized_size() as u64;
+                                
+                                // Read cost for entry point lookup in operations HashMap
+                                cost += model.cell_read(entry_point_size);
+                                
+                                // Conservative estimate for verifier key read during validation
+                                // We use a larger size to account for potentially large verifier keys
+                                // The actual size will be checked during application
```
tkerber: But... it's not? I thought the approach we'd agreed on was to use a _small_ stand-in initially, then do the read and use the real size. Neither seem to be happening.

ledger/src/verify.rs
```diff
@@ -1704,6 +1704,41 @@ where
...
+                        size: vk.serialized_size(),
+                        limit: crate::structure::MAX_VERIFIER_KEY_SIZE,
+                    });
+                }
+            }
+        }
+        
+        // Validate CMA keys size
+        let cma_size = self.initial_state.maintenance_authority.serialized_size();
+        const MAX_CMA_SIZE: usize = 8192; // 8 KiB max for CMA keys
```
tkerber: Not a fan of all the random constants that will be very hard to change. We have `TransactionLimits` for a reason.

tkerber (changes requested):

ledger/src/verify.rs
```diff
@@ -609,7 +616,7 @@ where
 
                 let mut fees = 0;
                 ref_state.param_check(true, |params| {
-                    fees = match self.fees(params, true) {
+                    fees = match self.fees_with_state(params, ref_state.ledger_state(), true) {
```
tkerber: This isn't safe – `param_check` will re-run iff the parameters changed. We need to factor this differently 😞

ledger/tests/contract_deployment.rs
```diff
@@ -42,7 +42,7 @@ async fn zero_contract_deploy_balance() {
 
     // Part 1: Deploy
     println!(":: Part 1: Deploy");
-    let count_op = ContractOperation::new(verifier_key(&RESOLVER, "count").await);
+    let count_op = ContractOperation::new(verifier_key(&RESOLVER, "count").await, None);
```
tkerber, line 45: It'd be nice if our tests also deployed the `.bzkir` files here, instead of defaulting to `None`. Not that it really makes a difference...

ledger/src/verify.rs
tkerber: My suggestion would be to instead add a `fees_with_impl` that is crate-pub, and add a new `StateReference` method `fees_check(FnOnce(Fn(ContractAddress, EntryPoint) -> ContractOperation) -> Result<()>) -> Result<()>`, that delegates to the appropriate contract, then we can still filter if that contract has changed or not.

tkerber (changes requested):
This appears not to address the concerns of limiting object size here – is that planned separately?

ledger/src/semantics.rs
```diff
@@ -1577,6 +1577,15 @@ impl<D: DB> LedgerState<D> {
                                 }
                             }
                         }
+                        let metadata_size = crate::verify::contract_metadata_size(&cstate);
```
tkerber: Imo this should be a per-entry-point limitation.

ledger/src/structure.rs
```diff
@@ -1232,6 +1232,8 @@ pub struct LedgerParameters {
     pub c_to_m_bridge_min_amount: u128,
     // The minimum value for `fee_prices.overall_price`.
     pub min_block_price: FixedPoint,
+    // A hard limit on the size of associated contract metadata.
+    pub max_contract_metadata_size: u64,
```
tkerber: Makes more sense in `TransactionLimits`?

ledger/src/verify.rs
```diff
@@ -96,8 +96,15 @@ pub trait StateReference<D: DB> {
...
     ) -> Result<(), MalformedTransaction<D>>;
+    fn fees_check(
+        &self,
+        check: impl FnOnce(
+            &dyn Fn(ContractAddress, &EntryPointBuf) -> Option<ContractOperation>,
+        ) -> Result<(), MalformedTransaction<D>>,
+    ) -> Result<(), MalformedTransaction<D>>;
     fn network_check(&self, network: &str) -> Result<(), MalformedTransaction<D>>;
     fn ref_state_hash(&self) -> ArenaHash<D::Hasher>;
+    fn ledger_state(&self) -> &LedgerState<D>;
```
tkerber: Is this used? This feels like a deliberate trapdoor to this abstraction; it's a smell to me.

tkerber (changes requested):

ledger/src/verify.rs
AlexShielded: It's vestigial. I'll remove it.

ledger/src/semantics.rs
tkerber: Ah, also, are we just enforcing it on maintenance updates here? I don't remember seeing a check at contract deploy, though I may just not be remembering it.

AlexShielded: I think it already is a per-entry point limitation...

```
   4 pub(crate) fn contract_metadata_size<D: DB>(state: &ContractState<D>) -> u64 {
   3     let mut size: u64 = 0u64;
   2     for entry: Sp<(Sp<EntryPointBuf, D>, …), …> in state.operations.iter() {
   1         size += entry.0.serialized_size() as u64;
390          size += entry.1.serialized_size() as u64;
   1     }
   2     size += state.maintenance_authority.serialized_size() as u64;
   3     size
   4 
```

AlexShielded: Added for `ContractAction::Deploy`

tkerber: > I think it already is a per-entry point limitation...

I'm confused. You code you show clearly shows it summing over all entry points, given a `ContractState`.

AlexShielded: Oh, I see - I'll fix that ASAP.

ledger/src/verify.rs
```diff
@@ -378,19 +374,21 @@ impl<D: DB> StateReference<D> for RevalidationReference<D> {
...
+pub(crate) fn check_entry_point_metadata_sizes<D: DB>(
+    state: &ContractState<D>,
+    limit: u64,
+) -> Result<(), (EntryPointBuf, u64)> {
+    let auth_size = state.maintenance_authority.serialized_size() as u64;
     for entry in state.operations.iter() {
-        size += entry.0.serialized_size() as u64;
-        size += entry.1.serialized_size() as u64;
+        let size =
+            entry.0.serialized_size() as u64 + entry.1.serialized_size() as u64 + auth_size;
```
tkerber: We probably want to test auth size independently, no?

tkerber (approved):

PR #450, opened 2026-04-22
perf(storage-core): wrap IrLoader key_to_child_repr in Rc

### LINKED ISSUE #449: `tagged_deserialize` on `MerkleTree` is O(n²) in node count
## Summary

`IrLoader::get` in `storage-core/src/arena.rs` holds `key_to_child_repr: HashMap<ArenaHash, ArenaKey>` by value. Each recursive call clones the map. The map has one entry per node in the graph being rebuilt, and an n-leaf `MerkleTree` drives ~2n recursive calls, so the whole deserialize comes out quadratic.

The map is built once and only read during recursion, so an `Rc` is enough — no interior mutability needed.

## Numbers

Criterion bench, height-32 `MerkleTree<(), InMemoryDB>`, n sequential leaves, `tagged_serialize` → `tagged_deserialize`:

| n      | time     | us/leaf |
|--------|----------|---------|
|    100 |   5.3 ms |   53    |
|  1,000 |  39.6 ms |   45    |
| 10,000 | 1986 ms  |  500    |

Per-leaf cost grows ~11× between n=1k and n=10k, which is the quadratic fingerprint. `serde_json::from_slice` on the same tree scales linearly, so the issue is specific to the tagged/Storable path.

## Fix

Wrap `key_to_child_repr` in `Rc` (matching `visited`). PR to follow.

After:

| n      | time    | us/leaf |
|--------|---------|---------|
|    100 |  2.6 ms |  26     |
|  1,000 | 25.8 ms |  26     |
| 10,000 |  260 ms |  26     |

Flat us/leaf — linear scaling restored.

geofflittle:
## Description

Closes #449.

`IrLoader::get` clones its `key_to_child_repr` HashMap on every recursive call. The map has one entry per node in the storage graph and an n-leaf `MerkleTree` takes ~2n recursive calls, so `tagged_deserialize` ends up O(n²). The map is built once and only read during recursion, so wrapping it in `Rc` is enough to avoid the per-call clone — no interior mutability needed. Restores linear scaling.

First commit adds a criterion bench across n=100/1k/10k at the Dust tree depth. Second is the fix.

Before / after:

| n      | before    | after   |
|--------|-----------|---------|
|    100 |   5.3 ms  |  2.6 ms |
|  1,000 |  39.6 ms  | 25.8 ms |
| 10,000 | **1986 ms** | **260 ms** |

us/leaf goes from growing (45 → 500 between n=1k and n=10k) to flat (~26 across all three).

No behavior change. `cargo +nightly test -p midnight-storage-core -p midnight-transient-crypto --lib` passes (37/37 + 55/55) on this branch.

## Sanity Checklist

This PR:

### DIFF
--- a/storage-core/src/arena.rs
+++ b/storage-core/src/arena.rs
@@ -913,7 +913,7 @@ impl<D: DB> Arena<D> {
                 .collect(),
             recursion_depth: recursive_depth,
             visited: Rc::new(RefCell::new(HashSet::new())),
-            key_to_child_repr,
+            key_to_child_repr: Rc::new(key_to_child_repr),
         }
         .get(&ArenaKey::Ref(key))?;
         if nodes == res.serialize_to_node_list() {
@@ -1085,7 +1085,7 @@ pub(crate) struct IrLoader<'a, D: DB> {
     recursion_depth: u32,
     /// The keys we've already deserialized once.
     visited: Rc<RefCell<HashSet<DynTypedArenaHash<D::Hasher>>>>,
-    key_to_child_repr: HashMap<ArenaHash<D::Hasher>, ArenaKey<D::Hasher>>,
+    key_to_child_repr: Rc<HashMap<ArenaHash<D::Hasher>, ArenaKey<D::Hasher>>>,
 }
 
 #[cfg(test)]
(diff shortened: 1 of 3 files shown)


tkerber (approved):

tkerber:
Thanks a lot! Well discovered!

tkerber:
Ah, this is blocked by commit signing. Please let me know if you want to fixup or if you want me to.

geofflittle:
Fixed-up!

tkerber (dismissed):

tkerber (approved):

PR #457, opened 2026-04-23
Expose commitmentTreeFirstFree and generatingTreeFirstFree on DustLocalState

jsidorenko:
### Summary

- Added getters to `commitment_tree_first_free` and `generating_tree_first_free` fields on `DustLocalState`
- Exposed both fields as read-only WASM getters under `commitmentTreeFirstFree` and `generatingTreeFirstFree` names

## Sanity Checklist

This PR:


ledger/src/dust.rs
```diff
@@ -1340,14 +1341,16 @@ impl Display for DustSpendError {
...
+#[derive(Debug, CopyGetters, Storable)]
 #[derive_where(Clone)]
 #[storable(db = D)]
 #[tag = "dust-local-state[v1]"]
 pub struct DustLocalState<D: DB> {
     pub generating_tree: MerkleTree<DustGenerationInfo, D>,
+    #[getset(get_copy = "pub")]
     generating_tree_first_free: u64,
     pub commitment_tree: MerkleTree<(), D>,
+    #[getset(get_copy = "pub")]
```
jacek-kurkowski-shielded: Why getset?
By looking at:
```
readonly generatingTreeFirstFree: bigint;
readonly commitmentTreeFirstFree: bigint;
```
it seems that you're only interested in getters

jsidorenko: `getset` is a macro (library's) name, I'm creating only getters here via the `get_copy = "pub"` param

jacek-kurkowski-shielded: Ahh, now I see, ignore my comment 👍

jacek-kurkowski-shielded (dismissed):

ledger/src/dust.rs
```diff
@@ -1340,14 +1341,16 @@ impl Display for DustSpendError {
...
 
-#[derive(Debug, Storable)]
+#[derive(Debug, CopyGetters, Storable)]
 #[derive_where(Clone)]
 #[storable(db = D)]
 #[tag = "dust-local-state[v1]"]
 pub struct DustLocalState<D: DB> {
     pub generating_tree: MerkleTree<DustGenerationInfo, D>,
+    #[getset(get_copy = "pub")]
     generating_tree_first_free: u64,
```
tkerber: Hmm... It does feel inconsistent to me that the trees are _fully_ public, but their corresponding `first_free` is restricted.

tkerber (approved):

PR #465, opened 2026-04-24
Fix panic in Node::remove when path is shorter than Extension's compressed_path

tkerber:
## Description

Add a bounds check before iterating over compressed_path in the Extension arm of Node::remove, matching the existing guard in lookup_with. Without this, a short path causes an index-out-of-bounds panic.

Also simplify the mismatch early-returns to use sp.clone() instead of reconstructing the node.

Note that this *in practice* can't be triggered due to most instances using fixed-length MPT keys, but should be resolved before we accidentally add a way to reach it.

## Sanity Checklist

This PR:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ with `zswap` being tracked in [Changelog Zswap](./CHANGELOG_zswap.md).
 
 # Change Log
 
+## Unreleased
+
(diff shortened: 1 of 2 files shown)


AlexShielded (approved):
LGTM

AlexShielded (changes requested):
Could we get a note in the changelog?

AlexShielded (approved):

PR #472, opened 2026-04-28
ZKIRv2 backwards compatibility

tkerber:
## Description

Extends the proposed optimisation PR with backwards-compatibility fully preserving old circuit behaviour. Most of this is tricky indirections to bypass the non-existence of the version field previously. Also adds this version field to zkir v3 and makes some parallel edits to avoid this much headache in the future.

## Sanity Checklist

This PR:

### DIFF
--- a/zkir/src/ir.rs
+++ b/zkir/src/ir.rs
@@ -23,7 +23,7 @@ use serde::{Deserialize, Serialize};
 #[cfg(feature = "proptest")]
 use serialize::randomised_serialization_test;
 use serialize::{Deserializable, Serializable, Tagged, tag_enforcement_test};
-use std::io::{self, Read};
+use std::io::{self, Read, Seek};
 use std::sync::Arc;
 use transient_crypto::curve::Fr;
 use transient_crypto::proofs::{
@@ -33,8 +33,10 @@ use transient_crypto::proofs::{
 /// A low-level IR allowing the prover to populate circuit witnesses.
 #[cfg_attr(feature = "proptest", derive(Arbitrary))]
 #[derive(Default, Clone, Debug, PartialEq, Serialize, Deserialize, Serializable)]
-#[tag = "ir-source[v2]"]
+#[tag = "ir-source[v2-generic]"]
 pub struct IrSource {
+    /// The minor version of this IR.
+    pub version: IrMinorVersion,
     /// The number of inputs, the initial elements in the memory
     pub num_inputs: u32,
     /// Whether or not this IR should compile a communications commitment
@@ -44,6 +46,53 @@ pub struct IrSource {
 }
 tag_enforcement_test!(IrSource);
 tag_enforcement_test!(ProverKey<IrSource>);
+
+#[derive(Serializable, Clone)]
+#[tag = "ir-source[v2]"]
+pub(crate) struct OldIrSource {
+    pub(crate) num_inputs: u32,
+    pub(crate) do_communications_commitment: bool,
+    pub(crate) instructions: Arc<Vec<Instruction>>,
+}
+
+impl From<OldIrSource> for IrSource {
+    fn from(value: OldIrSource) -> Self {
+        IrSource {
+            version: IrMinorVersion::V0,
+            num_inputs: value.num_inputs,
+            do_communications_commitment: value.do_communications_commitment,
+            instructions: value.instructions,
+        }
+    }
+}
+
+impl From<IrSource> for OldIrSource {
+    fn from(value: IrSource) -> Self {
+        OldIrSource {
+            num_inputs: value.num_inputs,
+            do_communications_commitment: value.do_communications_commitment,
+            instructions: value.instructions,
+        }
+    }
+}
+
+#[cfg_attr(feature = "proptest", derive(Arbitrary))]
+#[derive(
+    Clone, Debug, PartialEq, serde_repr::Serialize_repr, serde_repr::Deserialize_repr, Serializable,
+)]
+#[tag = "ir-minor-version[v2]"]
+#[repr(u8)]
+#[non_exhaustive]
+pub enum IrMinorVersion {
+    V0,
+    V1,
+}
+
+impl Default for IrMinorVersion {
+    fn default() -> Self {
+        IrMinorVersion::V1
+    }
+}
 
 impl Zkir for IrSource {
     fn check(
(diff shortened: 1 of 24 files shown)


zkir/src/ir.rs
```diff
@@ -45,6 +47,53 @@ pub struct IrSource {
...
+
+impl From<IrSource> for OldIrSource {
+    fn from(value: IrSource) -> Self {
+        OldIrSource {
+            num_inputs: value.num_inputs,
+            do_communications_commitment: value.do_communications_commitment,
+            instructions: value.instructions,
+        }
+    }
+}
```
miguel-ambrona: This should assert that `value.version` is `IrMinorVersion::V0`.

zkir/src/ir.rs
```diff
@@ -45,6 +47,53 @@ pub struct IrSource {
 tag_enforcement_test!(IrSource);
 tag_enforcement_test!(ProverKey<IrSource>);
 
+#[derive(Serializable, Clone)]
+#[tag = "ir-source[v2]"]
+pub(crate) struct OldIrSource {
+    pub(crate) num_inputs: u32,
+    pub(crate) do_communications_commitment: bool,
+    pub(crate) instructions: Arc<Vec<Instruction>>,
+}
```
miguel-ambrona, line 58: Do we really need to introduce this type? Wouldn't it be enough to assign the case `version: V0` of `IrSource` for it?

zkir/src/ir.rs
```diff
@@ -75,6 +124,32 @@ impl Zkir for IrSource {
...
     }
+
+    fn load_ir_from_tagged(reader: impl Read + Seek) -> io::Result<Self> {
+        Self::load_from_tagged(reader)
+    }
+
+    fn load_prover_key_from_tagged(mut reader: impl Read + Seek) -> io::Result<ProverKey<Self>> {
+        let pos = reader.stream_position()?;
+        // We try to first deserialize as the generic version. If this with an 'expected
+        // header tag' error, we rewind, and try again with the old tag.
```
miguel-ambrona: "If this with" -> "If this **fails** with"

zkir/src/ir.rs
```diff
@@ -75,6 +124,32 @@ impl Zkir for IrSource {
...
+        match serialize::tagged_deserialize::<ProverKey<IrSource>>(&mut reader) {
+            Ok(v) => return Ok(v),
+            Err(e) if !e.to_string().contains("expected header tag") => return Err(e),
+            _ => {}
+        }
```
miguel-ambrona: "expected header tag" is too general, shouldn't we be more specific about the actual header that was expected vs gotten?

A more robust approach would be to define a `peek_tag` function and then do this:
```
    if peek_tag(&mut reader)? == "prover-key[v7](ir-source[v2])" {
        // old fallback path
    } else {
        serialize::tagged_deserialize::<ProverKey<IrSource>>(&mut reader)
    }
```

zkir/src/ir.rs
```diff
@@ -386,11 +461,28 @@ impl IrSource {
...
+    /// - `ir-source[v2-generic]`
+    /// - `ir-source[v2]`
+    /// This is due to a need for a backwards-compatible format change.
+    pub fn load_from_tagged<R: Read + Seek>(mut reader: R) -> io::Result<Self> {
+        let pos = reader.stream_position()?;
+        // We try to first deserialize as the generic version. If this with an 'expected
+        // header tag' error, we rewind, and try again with the old tag.
+        match serialize::tagged_deserialize::<IrSource>(&mut reader) {
+            Ok(v) => return Ok(v),
+            Err(e) if !e.to_string().contains("expected header tag") => return Err(e),
```
miguel-ambrona: Same here, this is too general and fragile.

zkir/src/ir_vm.rs
```diff
@@ -836,10 +881,28 @@ impl Relation for IrSource {
...
+        } else {
+            // 0xff is a magic byte indicating non-v0 versions. These are tagged, and in a vec
+            // container. This byte is *not* representable in v0, as it is illegal for a SCALE
+            // encoded u32.
+            writer.write_all(&[0xff])?;
+            let mut raw = Vec::new();
+            tagged_serialize(&self, &mut raw)?;
+            raw.serialize(writer)
+        }
     }
```
miguel-ambrona, line 896: Another option would be to always write as in the `else` branch.

We only need to be general (and support old keys) when reading.

zkir/tests/proofs.rs
```diff
@@ -422,6 +422,7 @@ mod proof_tests {
            ]
         }"#;
         let ir = IrSource::load(ir_raw.as_bytes()).unwrap();
+        dbg!(&ir);
```
miguel-ambrona: leftover?

zkir/src/main.rs
```diff
@@ -124,6 +127,27 @@ fn extract_files_from_dir(dir: impl AsRef<Path>) -> anyhow::Result<Vec<OsString>
...
+            #[tag = "prover-key[v7](ir-source[v2])"]
+            struct FacadeProverKey(Vec<u8>);
+            let mut raw = Vec::new();
+            Serializable::serialize(pk, &mut raw)?;
+            let container = <Vec<u8>>::deserialize(&mut &raw[..], 0)?;
+            let facade = FacadeProverKey(container);
+            tagged_serialize(&facade, writer)
+        }
+        IrMinorVersion::V1 => tagged_serialize(pk, writer),
+        _ => unreachable!(),
```
miguel-ambrona: Do we need this last case?

Why did you make the version enum non-exhaustive?

zkir/src/main.rs
```diff
@@ -124,6 +127,27 @@ fn extract_files_from_dir(dir: impl AsRef<Path>) -> anyhow::Result<Vec<OsString>
...
+    match version {
+        IrMinorVersion::V0 => {
+            #[derive(Serializable)]
+            #[tag = "prover-key[v7](ir-source[v2])"]
+            struct FacadeProverKey(Vec<u8>);
+            let mut raw = Vec::new();
+            Serializable::serialize(pk, &mut raw)?;
+            let container = <Vec<u8>>::deserialize(&mut &raw[..], 0)?;
+            let facade = FacadeProverKey(container);
+            tagged_serialize(&facade, writer)
```
miguel-ambrona: This `FacadeProverKey` trick is duplicated in `ir.rs`. It could be good to unify it in a common helper.

miguel-ambrona (approved):
LGTM!

zkir/src/ir.rs
tkerber: You're right. Probably better as a `TryFrom`.

zkir/src/ir.rs
tkerber: We need the old serialization format. We could handroll it, but deriving it is easiest.

zkir/src/ir.rs
tkerber: Agreed, I think I avoided this because I'm not sure how tricky `peek_tag` would be to implement. Let me spike it...

zkir/src/ir_vm.rs
tkerber: Negative, because this would make old keys not reproducible, which isn't ideal. Imagine a contract you're using suddenly presents different keys, you might see that as a security issue. I'd like to avoid this.

zkir/src/main.rs
tkerber: non-exhausitive so adding a new version is formally not a breaking change (technically I guess this one _is_ a breaking change, but I'll pretend to ignore that). It makes sense for most consumers that they don't know of potential future versions that might get added.

annoyingly the binary isn't considered the same crate, so it can't exhaustively match a non-exhaustive marked enum.

zkir/src/main.rs
tkerber: Yeah, I wanted to not expose this in the public API, but exposing a pk-serialization function does make sense.

zkir/src/ir.rs
miguel-ambrona: Makes sense.

zkir/src/ir_vm.rs
miguel-ambrona: Fair enough.

miguel-ambrona (approved):
Good!

Shall we add a test to cover this specific versioning update?

tkerber:
> Shall we add a test to cover this specific versioning update?

I think I do want to add tests that ensure some existing IR compilation outputs remain unchanged. I've verified this by hand for now, but having it automated makes sense – it's inconveniently expensive though, so I want to mull over where that fits for a while...

PR #480, opened 2026-05-04
Add typed 'outputs' field to ZKIR v3

JosephDenman:
## Description

Added typed `outputs` field to ZKIR v3.

Sibling PR in `compact`: https://github.com/LFDT-Minokawa/compact/pull/421


zkir-v3/src/ir_vm.rs
```diff
@@ -584,6 +583,25 @@ impl IrSource {
...
+                bail!(
+                    "declared output {:?} has runtime type {:?} but signature declares {:?}",
+                    typed_id.name,
+                    value.get_type(),
+                    typed_id.val_t
+                );
+            }
+            for ir_val in encode_offcircuit(&value) {
+                outputs.push(ir_val);
+            }
```
miguel-ambrona: Why do you now push outputs in encoded form?

miguel-ambrona (commented):
I find the tests really verbose. If they are all positive tests, could we test several properties in the same circuit/test?

Otherwise, it may be time to start a module for ZKIR tests a bit more structured (in files).

zkir-v3/src/ir_vm.rs
JosephDenman: Because on line 620 we have this code block:

```
        if self.do_communications_commitment {
            let comm_comm = preimage
                .communications_commitment
                .ok_or(anyhow!("Expected communications randomness"))?;
            let mut comm_comm_inputs: Vec<Fr> = Vec::new();
            comm_comm_inputs.extend(preimage.inputs.iter());
            for output in outputs.iter() {
                comm_comm_inputs.push(output.clone().try_into()?);
            }
            if comm_comm.0 != transient_commit(&comm_comm_inputs[..], comm_comm.1) {
                error!(
                    ?comm_comm,
                    ?comm_comm_inputs,
                    "Communications commitment mismatch"
                );
                bail!("Communications commitment mismatch");
            }
        }
```

The `output.clone().try_into()?` call will fail if `output` is not an `IrValue::Native`.

miguel-ambrona: This is where I would call encode. It is more natural.

JosephDenman: Done

JosephDenman:
> Otherwise, it may be time to start a module for ZKIR tests a bit more structured (in files).

I've done this. Factored tests into different files and extracted common utilities.

miguel-ambrona (dismissed):
I understand `Instruction::Output` is now supposed to be the last instruction of a ZKIR program. What happens if it is not? Shall we produce an error or at least a warning?

A way to enforce this would be to redefine field "outputs" in the ZKIR metadata. Currently, it is used to declare the output types. Alternatively, "outputs" could simply be the variable identifiers (names). Note that we can infer the type of outputs (unlike inputs, where the type must be declared). That way we would not have an "Instruction::Output" and thus the issue I reported above would disappear.

zkir-v3/src/ir_vm.rs
miguel-ambrona: Perfect, thank you.

miguel-ambrona:
> > Otherwise, it may be time to start a module for ZKIR tests a bit more structured (in files).
> 
> I've done this. Factored tests into different files and extracted common utilities.

Great!
It would also be good to simplify/restructure the existing proofs.rs tests (break them in files). We can do that as we gradually implement the new ZKIR types.

JosephDenman:
> I understand Instruction::Output is now supposed to be the last instruction of a ZKIR program. What happens if it is not? Shall we produce an error or at least a warning?

I don't think the position of `Output` is an issue — a single `Output` mid-stream works as long as nothing after it adds to the outputs accumulator (which only `Output` itself does) and as long as the trailing instructions don't introduce inconsistencies the verifier rejects. The actual invariant is: `I::Output` must appear in `self.instructions` exactly once iff `self.outputs.is_empty()` is false, and exactly zero times otherwise. The current code only catches the violation indirectly via the `comm_comm` check, which only runs when `do_communications_commitment: true`. With `do_communications_commitment: false`, multiple or missing `Output`s slide through preprocess silently.

We will probably want a load time check of a number of different characteristics of ZKIR. I'll add a check a check for the above in a subsequent PR.

> A way to enforce this would be to redefine field "outputs" in the ZKIR metadata. Currently, it is used to declare the output types. Alternatively, "outputs" could simply be the variable identifiers (names). Note that we can infer the type of outputs (unlike inputs, where the type must be declared). That way we would not have an "Instruction::Output" and thus the issue I reported above would disappear.

This was my first approach, which I liked because of symmetry. Kevin says it complicates things with SSA and prefers the type only `outputs` field.

tkerber (approved):

PR #485, opened 2026-05-05
Implement symbolic Impact instructions for ZKIR v3

JosephDenman:
## Description

This PR is step toward cross-contract calls. The cross contract calls (CCC) work will require that we can execute the contents of `Impact` instructions (from ZKIR v3) against real contract states. The most intuitive way to achieve this is to represent Impact instructions symbolically instead of as (currently) an opaque lists of fields. This PR changes ZKIR v3 so that `Impact` instructions contain symbolic Impact VM ops. CCC itself ships in a follow-up.

### Changes

`Instruction::Impact` carries `Vec<ZkirOp>` instead of `Vec<Operand>`. `ZkirOp` is a new enum in `zkir-v3` that mirrors `onchain_vm::ops::Op`'s variant set with symbolic payloads — `Operand` references in place of `StateValue` / `Array<Key>` — and uses the same opcode bytes and field-element layout. `ZkirPushValue`, `ZkirKey`, and `ZkirReadResult` cover the operand-bearing positions.

`ir_vm.rs` walks `Vec<ZkirOp>` in both passes — the off-circuit preprocessor (`zkir_ops_to_field_elements_with_sizes`) emits the flat `Fr` stream, and the in-circuit synthesizer constrains the same encoding gated by the guard bit. The off-circuit pass also validates the prover-supplied `public_transcript_inputs` against what the encoded ops produce.

`IrType` gains an `alignment()` method used by the encoders.

### Non-Changes

Originally, I was going to make `Op` in `onchain-vm` generic to accommodate the symbolic payloads, but that would result in a change to tag decompositions, which I figured would be a non-starter. So, for now we take on the debt of maintaining a mirror op structure and keeping the two in sync.


zkir-v3/src/zkir_mode.rs
```diff
@@ -0,0 +1,689 @@
...
+    pub operands: Vec<Operand>,
+}
+
+/// ZKIR-mode `Op`. Mirrors `onchain_vm::ops::Op`'s variant set with symbolic
+/// payloads. Variant naming and serde encoding are kept consistent with the
+/// runtime so tests and Compact-emitted JSON read the same shape.
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, Serializable)]
+#[serde(rename_all = "lowercase", expecting = "operation")]
+#[tag = "zkir-op[v1]"]
+pub enum ZkirOp {
```
tkerber, line 98: This name is misleading imo. This is an impact operation, not a ZKIR one.

tkerber (commented):
I'm really not sure how I feel about this. I also think the direction is ultimately not my call, I'd ask @kmillikin to weigh in on if this is the right approach for ZKIR to take.

My concern is that this adds a really quite strong dependency from ZKIR onto Impact. I'm also not sure I agree with the assessment that this can't be exposed backwards-compatibly in `onchain-vm` – this may indeed be tricky, but mirroring the Impact behaviour seems to me like it might lead us to headaches down the line when we want to make Impact changes (e.g. modifications to the `log` op for events).

Let's say we have a new, breaking version of Impact – what effect will this have here?

miguel-ambrona (commented):
I do not understand the need for this. Can you explain it a bit more?

I wanted to get rid of the `Impact` instruction which I am not sure we need. This is making it significantly more involved, instead.

PR #486, opened 2026-05-06
Wasm constructors for DustStateChanges and ZswapStateChanges

jsidorenko:
## Description

This PR adds Wasm constructors for DustStateChanges and ZswapStateChanges classes

## Sanity Checklist

This PR:

### DIFF
--- a/ledger-wasm/ledger-v8.template.d.ts
+++ b/ledger-wasm/ledger-v8.template.d.ts
@@ -459,7 +459,7 @@ export class DustState {
 }
 
 export class DustStateChanges {
-  private constructor();
+  constructor(source: string, receivedUtxos: QualifiedDustOutput[], spentUtxos: QualifiedDustOutput[]);
   /**
    * The source of the state change, as a hex-encoded string
    */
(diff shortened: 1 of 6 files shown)


ledger-wasm/ledger-v8.template.d.ts
```diff
@@ -459,7 +459,8 @@ export class DustState {
 }
 
 export class DustStateChanges {
-  private constructor();
+  constructor(source: string, receivedUtxos: QualifiedDustOutput[], spentUtxos: QualifiedDustOutput[]);
```
tkerber: Please use `TransactionHash` over `string` here. I spent some time quite confused what this mysterious hex-encoded source string was 😅

ledger-wasm/ledger-v8.template.d.ts
```diff
@@ -1727,7 +1728,8 @@ export class ZswapChainState {
 }
 
 export class ZswapStateChanges {
-  private constructor();
+  constructor(source: string, receivedCoins: QualifiedShieldedCoinInfo[], spentCoins: QualifiedShieldedCoinInfo[]);
```
tkerber: Ditto

tkerber (changes requested):
Mostly good, just a nit to fix

ledger-wasm/ledger-v8.template.d.ts
jsidorenko: same :DD

tkerber (approved):

PR #495, opened 2026-05-11
Opaque Types in ZKIR v3

JosephDenman:
Add Opaque types in ZKIR v3:

Sibling PR in `compact`: https://github.com/LFDT-Minokawa/compact/pull/435


zkir-v3/src/ir_instructions/assign.rs
```diff
@@ -76,5 +76,17 @@ pub fn assign_incircuit(
             .jubjub()
             .assign_many(layouter, &convert_values::<JubjubFr>(values)?)
             .map(|xs| xs.into_iter().map(CircuitValue::JubjubScalar).collect()),
+
+        IrType::Opaque => {
+            // In-circuit, an Opaque input is just its commit witness.
```
miguel-ambrona, line 81: Rephrase this, "commit witness" is misleading.

"commit" -> "commitment"?

I would add more details. IIUC, an opaque type is represented in-circuit as the Poseidon hash of some payload. It is ok to use the word "commitment" since it is actually critical that this operation (from payload to in-circuit representation) be (at least computationally) binding. However, note that such commitment is not hiding, at least not in the strong sense, because it is deterministic. So I would be very clear in the relevant documentation about the fact that Opaque types DO NOT hide the payload.

zkir-v3/src/ir_instructions/assign.rs
```diff
@@ -76,5 +76,17 @@ pub fn assign_incircuit(
...
+            let fr_values = convert_values::<Fr>(values)?;
+            let field_values: Vec<Value<_>> =
+                fr_values.into_iter().map(|v| v.map(|fr| fr.0)).collect();
+            std_lib.assign_many(layouter, &field_values).map(|xs| {
+                xs.into_iter()
+                    .map(|commit| CircuitValue::Opaque { commit })
```
miguel-ambrona, line 87: In this context "commit" is a verb. The noun is "commitment".

zkir-v3/src/ir_types.rs
```diff
@@ -41,15 +40,69 @@ pub enum IrType {
...
+    /// preprocess pipeline and the in-circuit synthesis treat the
+    /// variable as a single-Fr value whose `Fr::try_from` returns
+    /// `commit` — meaning operands carrying Opaque values contribute
+    /// the cached commit Fr to the byte-flat impact stream's
+    /// `pis`/`public_transcript_inputs` exactly as if they were
+    /// `IrValue::Native(commit_fr)`. The preimage `bytes` survive only
+    /// at the IR-side memory layer for downstream consumers that
+    /// genuinely need the bytes (e.g. building runtime-VM `AlignedValue`s
+    /// from Compress preimages); the SNARK side never sees them.
+    #[serde(rename = "Opaque")]
```
miguel-ambrona, line 84: This massive comment is not aligned with the file style. Can it be more concise?

It also addresses some unimportant points like where Opaque appears. I would focus on what Opaque is, not on where it appears.

zkir-v3/src/ir_types.rs
```diff
@@ -65,14 +118,50 @@ pub enum IrValue {
...
+    /// The `commit` field is `transient_commit(&bytes,
+    /// Fr::from(bytes.len() as u64))`, computed once at construction time
+    /// (see [`IrValue::opaque`]). Subsequent `Fr::try_from(IrValue::Opaque
+    /// { commit, .. })` returns `commit` transparently, which is what the
+    /// off-circuit preprocess and in-circuit synthesis Impact arms read
+    /// for Compress-slot operands. The `bytes` field is preserved for
+    /// downstream IR-level consumers that need the preimage (e.g. when
+    /// reconstructing runtime-VM `AlignedValue`s for Compress slots);
+    /// the proving side itself only ever consumes `commit`.
+    Opaque { bytes: Vec<u8>, commit: Fr },
```
miguel-ambrona, line 132: commit -> commitment

zkir-v3/src/ir_types.rs
```diff
@@ -41,15 +40,69 @@ pub enum IrType {
...
     pub fn encoded_len(&self) -> usize {
         match self {
             IrType::Native => 1,
             IrType::JubjubPoint => 2,
             IrType::JubjubScalar => 1,
+            IrType::Opaque => panic!(
+                "IrType::Opaque has variable encoded length; callers must \
+                 read byte_len explicitly. See IrSource::preprocess for the \
+                 canonical pattern."
+            ),
```
miguel-ambrona, line 105: We should not panic here and return 1 instead, right?

zkir-v3/src/ir_types.rs
```diff
@@ -93,22 +186,25 @@ pub enum CircuitValue {
...
 
 impl CircuitValue {
-    pub fn value(&self) -> Value<IrValue> {
-        match self {
-            CircuitValue::Native(x) => x.value().cloned().map(|x| IrValue::Native(Fr(x))),
-            CircuitValue::JubjubPoint(p) => p.value().map(IrValue::JubjubPoint),
-            CircuitValue::JubjubScalar(s) => s.value().map(IrValue::JubjubScalar),
-        }
-    }
-
```
miguel-ambrona, line 106: Why did you remove this?

zkir-v3/src/ir_types.rs
```diff
@@ -144,20 +240,210 @@ macro_rules! impl_enum_from_try_from {
...
 
-// Derives implementations, for every basic type T:
-//  - From<T> for IrValue
-//  - TryFrom<IrValue> for T
+// Derives implementations for non-Native variants. We don't use the macro
+// for `Native => Fr` because we want `Fr::try_from(IrValue)` to succeed for
+// both `Native` (returning the inner Fr) and `Opaque` (returning the
+// precomputed commit) — the macro can only produce one impl per
+// (variant, type) pair, so the Native + Opaque dispatch is written manually
+// below.
```
miguel-ambrona, line 248: Please, keep the original comment and add an extra note after it.

miguel-ambrona (commented):
I'm skeptical about these changes.

I particularly dislike the fact that Opaque decoding is possible off-circuit but not in-circuit. IIUC, that means we cannot use `Instruction::Decode` (or encode) in a ZKIR program (because it would panic in the in-circuit execution). In that case, why does the off-circuit implementation exist at all?

Do we really need to introduce opaque types in ZKIR?
Can't the payload of opaque types be calculated in the off-circuit pass prior to ZKIR (the one used to create `ProofPreimage`), and stored in the `ProofPreimage`?

zkir-v3/src/ir_instructions/encode.rs
```diff
@@ -26,7 +26,15 @@ use crate::{
 };
 
 /// Encodes the given off-circuit value as a vector of IrValue::Native.
+///
+/// For `IrValue::Opaque { bytes, .. }` the output is the canonical
+/// runtime input-vector encoding `[byte_len, preimage_fr_0, ...,
+/// preimage_fr_{N-1}]` (each entry wrapped in `IrValue::Native`). This
+/// matches what the JS bridge flattens into `ProofPreimage.inputs` for
+/// an Opaque-typed input and what `decode_offcircuit(_, &IrType::Opaque)`
+/// reconstructs.
```
iquerejeta, line 35: IMO this comment should be in the `Opaque` branch, not at the function level.

zkir-v3/src/ir_instructions/encode.rs
```diff
@@ -43,6 +51,26 @@ pub fn encode_offcircuit(value: &IrValue) -> Vec<IrValue> {
...
+                .map(|chunk| {
+                    Fr::from_le_bytes(chunk)
+                        .expect("Opaque preimage chunk must fit into Fr")
+                        .0
+                })
+                .rev()
+                .collect();
+            frs.extend(packed);
+            frs
+        }
```
iquerejeta, line 73: What is this representation used for? This might be used to declare public inputs, for which we should run the commitment function.

zkir-v3/src/ir_instructions/encode.rs
```diff
@@ -72,6 +100,88 @@ pub fn encode_incircuit(
             assert_eq!(encoded.len(), 1);
             Ok(encoded)
         }
+        CircuitValue::Opaque { .. } => {
+            return Err(Error::Synthesis(
+                "encode_incircuit: CircuitValue::Opaque encoding is not implemented; \
+                 the in-circuit projection currently carries only the commit, not \
+                 the preimage Frs that the off-circuit `encode_offcircuit` emits."
+                    .into(),
+            ));
```
iquerejeta, line 109: Why no encode opaque types in-circuit as their commitment?

zkir-v3/src/ir_instructions/encode.rs
```diff
@@ -72,6 +100,88 @@ pub fn encode_incircuit(
...
+///
+/// For [`IrValue::Opaque`] this emits exactly the cached `commit` Fr,
+/// matching the runtime's `<ValueAtom as ValueAtomExt>::field_repr_unchecked`
+/// `Compress` arm at `transient-crypto/src/fab.rs` (and its
+/// empty-AV special case that writes `Fr::from(0u64)`). Because
+/// `IrValue::opaque` precomputes `commit` from the same `transient_commit(
+/// bytes, Fr::from(byte_len as u64))` formula, the IR-side comm_comm
+/// preimage built from these emissions matches the JS bridge's
+/// commit-bearing flatten over the same `AlignedValue` Compress slots.
+pub fn encode_offcircuit_for_commit(value: &IrValue) -> Vec<IrValue> {
```
iquerejeta, line 133: I see that you've added these two additional functions, `encode_offcircuit_for_commit` and `encode_incircuit_for_commit`. I imagine that using directly `encode_incircuit/offcircuit` did not work at some point - could you point out what the issue was?

zkir-v3/src/ir_instructions/encode.rs
iquerejeta: Below you mention 'The preimage `bytes` survive only at the IR-side', and the bytes are used as bytes, so I don't understand why you would need this representation as field elements. 

This packed byte representation makes sense for bytes, as we want to represent them as public inputs as concisely as possible. However, I don't see the point for opaque types, as they already have a concise representation in-circuit.

zkir-v3/src/ir_types.rs
```diff
@@ -65,14 +118,50 @@ pub enum IrValue {
 
     /// Jubjub scalar field value.
     JubjubScalar(JubjubFr),
+
+    /// Opaque preimage carried alongside its `transient_commit` commitment.
+    /// The `commit` field is `transient_commit(&bytes,
+    /// Fr::from(bytes.len() as u64))`, computed once at construction time
+    /// (see [`IrValue::opaque`]). Subsequent `Fr::try_from(IrValue::Opaque
+    /// { commit, .. })` returns `commit` transparently, which is what the
```
iquerejeta, line 126: The linter inserted line breaks inside backtick-delimited text.

zkir-v3/src/ir_types.rs
```diff
@@ -41,15 +40,69 @@ pub enum IrType {
...
+    Opaque,
 }
 
 impl IrType {
     /// Number of raw `Fr` elements needed to represent a value of this type.
+    ///
+    /// Panics for `IrType::Opaque`, which is variable-length at runtime —
+    /// callers that need to slice an Opaque from an `Fr` stream must read
+    /// the leading `byte_len` Fr first and consume `1 + ceil(byte_len /
+    /// FR_BYTES_STORED)` Frs total. See the per-type slicer in
```
iquerejeta, line 94: The linter inserted line breaks inside backtick-delimited text.

zkir-v3/src/ir_types.rs
iquerejeta: There's a few places else where this happens

iquerejeta (commented):
Do we have a spec of how we expect an Opaque type to behave in/off-circuit? Maybe some of the comments to this PR are due to not really understanding they role and usage.

iquerejeta (commented):
Do we have a spec of how we expect an Opaque type to behave in/off-circuit? Maybe some of the comments to this PR are due to not really understanding they role and usage.

zkir-v3/src/ir_types.rs
JosephDenman: `CircuitValue::value()` had a single caller: the value-equality cross-check in `mem_insert` that was on line 723. That check is the one the base code itself flagged as unnecessary, so I replaced it with a guard — each IR variable must be bound at most once during synthesize — which catches the real "prepare/synthesize disagree" bug class. That left `value()` with no callers.

PR #498, opened 2026-05-11
Add ECDSA signature support

tkerber:
## Description

Adds ECDSA signatures for unshielded tokens & CMAs.

## Sanity Checklist

This PR:
- [X] contains changes to architecture [^1]
- [X] contains breaking API changes [^1][^2]


onchain-runtime-wasm/src/primitives.rs
```diff
@@ -66,35 +67,74 @@ pub fn communication_commitment(
...
-pub fn signature_verifying_key(key: &str) -> Result<String, JsError> {
-    let key: schnorr::SigningKey = from_value_hex_ser(key)?;
-    to_value_hex_ser(&key.verifying_key())
+pub fn signature_verifying_key(sk: JsValue) -> Result<JsValue, JsError> {
+    let sk: PreSignature = from_value(sk)?;
+    let vk = match sk {
+        PreSignature::Schnorr(raw) => PreSignature::Schnorr(to_value_hex_ser(
+            &from_value_hex_ser::<schnorr::SigningKey>(&raw)?.verifying_key(),
+        )?),
+        PreSignature::ECDSA(raw) => PreSignature::Schnorr(to_value_hex_ser(
```
kapke: `PreSignature::ECDSA`?

spec/preliminaries.md
```diff
@@ -56,14 +56,18 @@ const NIGHT_RAW: RawTokenType = [0u8; 32];
...
+
+- Schnorr over Secp256k1, as specified in [BIP 340](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki).
+- ECDSA over Secp256k1
+
+Signing keys, verifying keys, and signatures are disjoint unions over these algorithms.
 
 ```rust
-type SigningKey = secp256k1::Scalar;
-type VerifyingKey = secp256k1::Point;
+type SigningKey = ...
```
kapke: Would it make sense to expand on the details here, or the paragraph above?

onchain-runtime-wasm/src/primitives.rs
```diff
@@ -66,35 +67,74 @@ pub fn communication_commitment(
...
+            &schnorr::SigningKey::sample(OsRng),
+        )?))?,
+        Some("ecdsa") => to_value(&PreSignature::ECDSA(to_value_hex_ser(
+            &ecdsa::SigningKey::sample(OsRng),
+        )?))?,
+        Some(kind) => return Err(JsError::new(&format!("Unknown signature kind: {kind}"))),
+    })
 }
 
 #[wasm_bindgen(js_name = "signingKeyFromBip340")]
```
kapke, line 82: Should a similar function be created for ECDSA for completeness?

kapke:
Test coverage is really needed, so that there is an evidence that the plumbing covers the desired flows and is complete.

onchain-runtime-wasm/src/primitives.rs
tkerber: 🤦 I knew I'd mess up one of these...

onchain-runtime-wasm/src/primitives.rs
tkerber: Good question. The naming scheme here suggested this has a specific relation to the BIP-340 standard though. I'm not sure what the equivalent should be for ECDSA.

I'm not even sure if this technically does more than just hex-encoding its input 🤔

spec/preliminaries.md
tkerber: Maybe... Although I'm not sure _what_ details to add.

ledger/tests/maintenance.rs
```diff
@@ -130,10 +130,13 @@ fn maintenance() {
     strictness.enforce_balancing = false;
     let fake_vk = VerifierKey::deserialize(&mut &b"\x00\x00\x00\x00"[..], 0).unwrap();
 
-    let committee_sks: Vec<_> = (0..4).map(|_| SigningKey::sample(&mut rng)).collect();
+    let committee_sks: Vec<_> = (0..4)
+        .map(|_| base_crypto::schnorr::SigningKey::sample(&mut rng))
+        .collect();
```
iquerejeta, line 135: Should we add tests for ECDSA as well?

onchain-runtime-wasm/src/primitives.rs
```diff
@@ -66,35 +67,74 @@ pub fn communication_commitment(
...
-pub fn signature_verifying_key(key: &str) -> Result<String, JsError> {
-    let key: schnorr::SigningKey = from_value_hex_ser(key)?;
-    to_value_hex_ser(&key.verifying_key())
+pub fn signature_verifying_key(sk: JsValue) -> Result<JsValue, JsError> {
+    let sk: Signatureish = from_value(sk)?;
+    let vk = match sk {
+        Signatureish::Schnorr(raw) => Signatureish::Schnorr(to_value_hex_ser(
+            &from_value_hex_ser::<schnorr::SigningKey>(&raw)?.verifying_key(),
+        )?),
+        Signatureish::ECDSA(raw) => Signatureish::Schnorr(to_value_hex_ser(
```
iquerejeta: Invalid mapping

onchain-runtime-wasm/src/conversions.rs
```diff
@@ -23,6 +23,15 @@ use js_sys::Uint8Array;
 use serde::{Deserialize, Serialize};
 use wasm_bindgen::prelude::*;
 
+#[derive(Serialize, Deserialize)]
+#[serde(tag = "tag", content = "value")]
+#[serde(rename_all = "kebab-case")]
+pub enum Signatureish {
```
iquerejeta: This type seems to represent VKs, SKs and Signatures. Don't know if that's a common technique for JS <> Rust bindigs, but if so I would add comment /documentation in the enum to explain what it is. And maybe a different name? e.g. SignatureMaterial

iquerejeta: Oh, I see you rename it - maybe still add docs?

spec/preliminaries.md
```diff
@@ -56,14 +56,18 @@ const NIGHT_RAW: RawTokenType = [0u8; 32];
 
 ## Signatures
 
-We also need to assume public key cryptography. We use Schnorr over Secp256k1,
-as specified in [BIP 340](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki).
+We also need to assume public key cryptography. We support multiple algorithms, currently consisting of:
+
+- Schnorr over Secp256k1, as specified in [BIP 340](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki).
+- ECDSA over Secp256k1
```
iquerejeta: We should specify `ECDSA-SHA256 over secp256k1`. Also, an important note is that ECDSA is malleable, in the sense that a valid signature can be modified by someone different than the signer. The resulting signature is valid for the same message, but this could be an issue if we somehow rely on transaction non-malleability.

iquerejeta: Checked, and indeed we are rejecting signatures with `high_s`. Therefore, the ECDSA signatures we accept as valid are not malleable. 

https://github.com/shieldedtech/product/pull/101#discussion_r3241885351

onchain-runtime-wasm/src/primitives.rs
tkerber: It is indeed just hex encoding things. I think it's therefore not needed / I'd rather lean to removing the `FromBip340`, though I'm not in a rush for that.

onchain-runtime-wasm/src/conversions.rs
tkerber: Added a comment. It's really just a serialization wrapper.

spec/preliminaries.md
tkerber: Thanks for the clarification. I changed to `ECDSA-SHA256 over secp256k1 with high-s rejection`, which is hopefully specific enough.

spec/preliminaries.md
tkerber: I expanded this out with the variants, although I'm not really inclined to adding the details of the key/signature structures of each scheme.

iquerejeta (dismissed):
LGTM

kapke (dismissed):
LGTM!

AlexShielded (dismissed):
LGTM!

jacek-kurkowski-shielded (dismissed):
💯

kapke (approved):

AlexShielded (approved):

gilescope (approved):

jacek-kurkowski-shielded (approved):

PR #500, opened 2026-05-12
Allow setting dust sync time

jsidorenko:
## Description

Currently, the dust sync time could only be set via the processed events.
This PR allows setting it manually, which is required by the event-less sync model.

## Sanity Checklist

This PR:

### DIFF
--- a/ledger-wasm/src/dust.rs
+++ b/ledger-wasm/src/dust.rs
@@ -1586,6 +1586,13 @@ impl DustLocalState {
     pub fn sync_time(&self) -> Date {
         seconds_to_js_date(self.0.sync_time.to_secs())
     }
+
+    #[wasm_bindgen(setter, js_name = "syncTime")]
+    pub fn set_sync_time(&mut self, sync_time: &Date) -> Result<(), JsError> {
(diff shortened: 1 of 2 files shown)


ledger-wasm/src/dust.rs
```diff
@@ -1586,6 +1586,13 @@ impl DustLocalState {
     pub fn sync_time(&self) -> Date {
         seconds_to_js_date(self.0.sync_time.to_secs())
     }
+
+    #[wasm_bindgen(setter, js_name = "syncTime")]
```
jacek-kurkowski-shielded, line 1590: `setSyncTime` ?

jacek-kurkowski-shielded (changes requested):
Shouldn't it be named `setSyncTime`?

ledger-wasm/src/dust.rs
jsidorenko: nope, for setters we use the same js_name as for the field getter

jacek-kurkowski-shielded: Ohh, indeed, my bad!

jacek-kurkowski-shielded (approved):
LGTM!

PR #506, opened 2026-05-15
feat: infallible post_block_update

Klapeyron:
## Description

Separation of fallible and infallible parts of `post_block_update`, so validation can be triggered before `post_block_update` is executed making it infallible at the same time

## Sanity Checklist

This PR:

### DIFF
--- a/ledger/src/semantics.rs
+++ b/ledger/src/semantics.rs
@@ -1532,13 +1532,11 @@ impl<D: DB> LedgerState<D> {
     }
 
     #[instrument(skip(self))]
-    pub fn post_block_update(
+    pub fn prevalidate_post_block_update(
         &self,
-        tblock: Timestamp,
         detailed_block_fullness: NormalizedCost,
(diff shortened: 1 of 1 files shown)


ledger/src/semantics.rs
```diff
@@ -1558,6 +1556,21 @@ impl<D: DB> LedgerState<D> {
...
         }
+
+        Ok(())
+    }
+
+    #[instrument(skip(self))]
+    /// Applies fee-price adjustments and submodule `post_block_update` hooks.
+    ///
+    /// Caller must ensure [`prevalidate_post_block_update`](Self::prevalidate_post_block_update) succeeds first.
+    pub fn apply_post_block_update(
```
Klapeyron, line 1567: If that will be fine, I will prefer to keep it public for now, so I can have more flexibility in the node to either use prevalidate and pure apply later or prevalidate followed by fallible post_block_update

tkerber (approved):

PR #522, opened 2026-05-21
fix: changed force_as_arc lock hold logic

### LINKED ISSUE #521: fix(storage-core): runtime-park hang in Sp::force_as_arc / BackendLoader::get under concurrent collapsed-update load
### Summary

Under heavy concurrent merkle-tree-update load, `Sp::force_as_arc` (called via `BackendLoader::get` from `make_dust_commitments_collapsed_update`, `make_dust_generations_collapsed_update`, and related collapsed-update paths) parks the entire tokio runtime indefinitely. The pod produces zero log output (no errors, no panics, no background heartbeats) until kubelet's `/live` probe fails and the container is SIGTERM'd.

This is **distinct from the TOCTOU race fixed in [PR #499](https://github.com/midnightntwrk/midnight-ledger/pull/499) / `storage-core-1.2.1-rc.1`**, which closed a different surface symptom (panic at `arena.rs:811`: `attempted to increment non-existant ref`). PR #499 is necessary but not sufficient: the runtime-park hang reproduces on the indexer test image built against `storage-core-1.2.1-rc.1`. Same code family, different surface symptom.

### Evidence — test image deploy (20 May 2026)

Indexer test image built with all midnight-ledger family crates patched to `storage-core-1.2.1-rc.1` tag (commit `f3bc591f`). Tag: `v4.3.2-rc.3-tomasz-ledger-test`. Deployed to qanet-blue with `RUST_LOG=debug` and the heavy combined query below at parallelism=200.

Query:
```graphql
query {
  d1: dustGenerationMerkleTreeUpdate(startIndex: 0, endIndex: 1024) { startIndex endIndex update protocolVersion }
  d2: dustCommitmentMerkleTreeUpdate(startIndex: 0, endIndex: 64)   { startIndex endIndex update protocolVersion }
  z:  zswapMerkleTreeCollapsedUpdate(startIndex: 0, endIndex: 128)  { startIndex endIndex update protocolVersion }
  b:  block { height hash transactions { hash raw protocolVersion } }
}
```

**Smoking-gun log line** — last output from a pod before kubelet kill (exit 143):

```
2026-05-20T21:32:02.574146Z DEBUG indexer_api::domain::ledger_state
  file=indexer-api/src/domain/ledger_state.rs line=169
  message="creating dust commitments collapsed update"
  end_index=64 start_index=0
  trace_id=715594bd65c52edfca782f822160cd53
```

Then **zero output for ~58 seconds**. `indexer_api::application`'s background "block indexed" trace (normally fires every ~6 seconds as the chain progresses) stopped immediately after the line above. No tokio scheduler interleaving, no error, no panic. The runtime is fully parked.

All 4 indexer-api pods restarted 2–4 times each via kubelet `/live` SIGTERM over the test window. Failure rate (`HTTP 5xx + connect failures`) was 44% under debug logging, 4% without (debug overhead shifts race-window timing).

### Why it looks like a lock-ordering issue (not CPU starvation)

The pod has CPU resources available — block-indexed events are scheduled by a different tokio task than request handlers. If the issue were plain CPU starvation, those events would still emit (just slower). Their complete silence alongside the request-handler silence indicates a thread holding a `parking_lot::ReentrantMutex` (metadata / sp_cache / backend) is itself parked, blocking all other tokio workers that need to acquire any of those mutexes.

This matches Thomas Kerber's earlier description:
> "tokio runtime fully parked with `parking_lot::ReentrantMutex` (metadata / sp_cache / backend) held indefinitely, indexer-api pool sitting at 199-200 of 200 connections frozen idle for hours."

### Suspected root cause

The existing `force_as_arc_lock_ordering_regression` test (around `storage-core/src/arena.rs:2786`) catches one ABBA pattern between `force_as_arc` and `new_sp_locked`. The runtime-park hang likely involves a **sibling lock-ordering inversion** reachable through the `BackendLoader::get` call chain that the existing test does not cover.

Specifically, `make_dust_commitments_collapsed_update` (and the dust-generations / zswap collapsed-update variants) walks a tree where each `.deref()` triggers `force_as_arc -> from_arena -> BackendLoader::get` recursively across child `Sp`s. With multiple concurrent walks of overlapping tree paths from separate tokio worker threads, the recursive lock acquisition order varies based on which Sp is hit first.

### Reproducer status

- **PR #499 surface (panic)**: Reproducer at [`midnightntwrk/midnight-ledger@15259cf3`](https://github.com/midnightntwrk/midnight-ledger/commit/15259cf3d1b755dd301a7ae25cf848e39db48cb5) fires reliably (3/3 runs, 7-of-8-thread panic at `arena.rs:811`). Passes on `storage-core-1.2.1-rc.1`.
- **This ticket's surface (runtime hang)**: No deterministic reproducer yet. Currently reproducible only via live qanet load:
  - Giuseppe's WS regression suite at high parallelism (Slack thread, 20 May: 70% pass rate at "max parallelism", 100% at "5 concurrent threads")
  - HTTP load at 200 parallel with the combined-aliases query above (qanet-blue test image, 20 May)

Giles Cope suggested [`tokio-rs/loom`](https://github.com/tokio-rs/loom) for systematic concurrency-permutation testing as a path to a deterministic reproducer. Noted but not attempted yet.

### Why it only started appearing recently

Working hypothesis (per Thomas): the recent qanet SRE cost-reduction PR (shielded-gitops #2088) changed the per-pod resource envelope. The race condition appears to have been latent and is now surfaced by the new concurrency profile under cold-cache load on tighter resources. Cold-start pods (cycled by gitops/ArgoCD) deadlock under QA load; manually-restarted pods are much more resilient (~17 hour clean runway observed in one case).

### Indexer-side mitigation (already shipped)

[indexer PR #1145](https://github.com/midnightntwrk/midnight-indexer/pull/1145) added a `/live` HTTP endpoint at indexer-api, wired to kubelet's liveness probe. When a pod deadlocks, kubelet detects the timeout and restarts the container (~90s recovery). This converts what was a hours-of-manual-intervention incident into a transient outage and is in production on qanet from 4.3.2-rc.1 onward.

The structural fix (this ticket) eliminates the underlying race; the indexer mitigation handles recovery if the race fires.

### Suggested investigation

1. Audit `storage-core/src/arena.rs` around `Sp::force_as_arc`, focusing on the lock acquisition order through `from_arena -> BackendLoader::get` recursion across child `Sp`s. The existing `force_as_arc_lock_ordering_regression` test covers `force_as_arc <-> new_sp_locked`; check sibling inversions involving `metadata` / `sp_cache` / `backend` interactions reachable from `BackendLoader::get`.
2. Write a loom-based test (sketch from Giles) targeting multi-thread + multi-key + nested-Sp paths. Sean's existing reproducer (8 OS threads, 200 iterations, 10 roots, nested lazy struct) is a starting point for the threading shape.
3. Consider whether the `parking_lot::ReentrantMutex` choice is contributing — `parking_lot` ReentrantMutex doesn't track reentrancy depth via a fairness mechanism; if a thread re-enters while a sibling lock acquisition is in flight, recovery from contention is order-of-arrival rather than fairness-bounded.

### Acceptance

- A failing test (loom-based or otherwise) that reproduces the runtime-park hang on `storage-core-1.2.1-rc.1`
- The fix passes both the new test and Sean's existing reproducer ([`15259cf3`](https://github.com/midnightntwrk/midnight-ledger/commit/15259cf3d1b755dd301a7ae25cf848e39db48cb5))
- Validation on indexer test image (Giuseppe's WS regression suite at high parallelism against `indexer-blue.qanet.midnight.network` with the fixed image deployed)

### Related

- [PR #499](https://github.com/midnightntwrk/midnight-ledger/pull/499) (storage-core 1.2.1-rc.1, TOCTOU panic surface fix — necessary but not sufficient for this ticket)
- [midnight-ledger#509](https://github.com/midnightntwrk/midnight-ledger/issues/509) (regression coverage for the TOCTOU surface — separate workstream)
- [midnight-indexer#1155](https://github.com/midnightntwrk/midnight-indexer/issues/1155) (indexer-side bump of storage-core, deferred pending this ticket's resolution)
- [midnight-indexer#1145](https://github.com/midnightntwrk/midnight-indexer/pull/1145) (kubelet `/live` mitigation, already shipped)
- gitops PR for the test image deploy: [shielded-gitops#2132](https://github.com/shieldedtech/shielded-gitops/pull/2132)
- Indexer test image tag: [`v4.3.2-rc.3-tomasz-ledger-test`](https://github.com/midnightntwrk/midnight-indexer/releases/tag/v4.3.2-rc.3-tomasz-ledger-test)
- Slack threads in `topic-sre` / `topic-qa` from 13–20 May 2026

Klapeyron:
## Description

https://github.com/midnightntwrk/midnight-ledger/pull/262 & https://github.com/midnightntwrk/midnight-ledger/pull/499 fixed the ABBA lock-ordering issue by acquiring locks in `metadata -> sp_cache` order, but it held those locks for too long.

On a cache miss, `force_as_arc` called `Sp::from_arena(...)` while still holding both `metadata` and `sp_cache`. That path goes through `BackendLoader::get`, backend access, deserialization, and child `Sp` loading.

Under concurrent collapsed Merkle update walks, one Tokio worker could sit inside that slow or recursive load while holding arena locks, causing other workers to pile up behind those locks. That matches the observed “no panic, no logs, runtime appears parked” symptom.
## Sanity Checklist

This PR:

### DIFF
--- a/CHANGELOG_storage_core.md
+++ b/CHANGELOG_storage_core.md
@@ -11,6 +11,7 @@
 - fix: remove pending Update from memory before cache_insert_new_key in get()
 - fix: Respect lock ordering in `force_as_arc`
 - fix: hold metadata lock across `track_lazy` and `Sp::lazy` in `BackendLoader::get` lazy path to prevent a panic when a concurrent drop removes the ref-counted entry between the two calls
+- fix: avoid holding arena metadata and `Sp` cache locks while `force_as_arc` loads from the backend
 
 ## Version `1.1.0`
 
--- a/storage-core/src/arena.rs
+++ b/storage-core/src/arena.rs
@@ -976,14 +976,9 @@ impl<D: DB> Loader<D> for BackendLoader<'_, D> {
             let sp = {
                 let metadata_lock = self.arena.lock_metadata();
                 if let ArenaKey::Ref(key) = child {
-                    self.arena
-                        .track_lazy(&metadata_lock, key.clone(), child);
+                    self.arena.track_lazy(&metadata_lock, key.clone(), child);
                 }
-                Sp::lazy(
-                    self.arena.clone(),
-                    child.hash().clone(),
-                    child.clone(),
-                )
+                Sp::lazy(self.arena.clone(), child.hash().clone(), child.clone())
             };
             return Ok(sp);
         }
@@ -1637,18 +1632,12 @@ impl<T: Storable<D>, D: DB> Sp<T, D> {
     fn force_as_arc(&self) -> &Arc<T> {
         // Initialize `OnceLock` if necessary.
         if self.data.get().is_none() {
-            // Acquire metadata before sp_cache to respect the documented lock
-            // ordering (metadata → sp_cache → backend; see Arena struct).
-            // from_arena and its callees (track_lazy, increment_ref) also
-            // acquire metadata, but that is safe because the locks are
-            // reentrant. Acquiring sp_cache first would invert the ordering
-            // and deadlock against new_sp_locked (which holds metadata, then
-            // acquires sp_cache).
-            let _metadata_lock = self.arena.lock_metadata();
-            let cache_lock = self.arena.lock_sp_cache();
-            let maybe_arc = self
-                .arena
-                .read_sp_cache_locked::<T>(&cache_lock, &self.root);
+            let maybe_arc = {
+                let _metadata_lock = self.arena.lock_metadata();
+                let cache_lock = self.arena.lock_sp_cache();
+                self.arena
+                    .read_sp_cache_locked::<T>(&cache_lock, &self.root)
+            };
             let arc: Arc<T> = match maybe_arc {
                 Some(arc) => arc,
                 None => {
@@ -1665,23 +1654,20 @@ impl<T: Storable<D>, D: DB> Sp<T, D> {
                                 std::any::type_name::<T>()
                             ),
                         };
-                    let arc = sp
-                        .data
+                    sp.data
                         .take()
-                        .expect("result of Sp::from_arena should be initialized");
-                    if let ArenaKey::Ref(_) = &self.child_repr {
-                        self.arena.write_sp_cache_locked(
-                            &cache_lock,
-                            self.root.clone(),
-                            arc.clone(),
-                        );
-                    }
-                    arc
+                        .expect("result of Sp::from_arena should be initialized")
                 }
             };
             // We don't care if this succeeds: failure just means
             // someone else set the same value in another thread.
             let _ = self.data.set(arc);
+            if let (ArenaKey::Ref(_), Some(arc)) = (&self.child_repr, self.data.get()) {
+                let _metadata_lock = self.arena.lock_metadata();
+                let cache_lock = self.arena.lock_sp_cache();
+                self.arena
+                    .write_sp_cache_locked(&cache_lock, self.root.clone(), arc.clone());
+            }
         }
         self.data.get().unwrap()
     }


tkerber:
:-1: I'm not changing core storage logic on vibes.

> Under concurrent collapsed Merkle update walks, one Tokio worker could sit inside that slow or recursive load while holding arena locks, causing other workers to pile up behind those locks. That matches the observed “no panic, no logs, runtime appears parked” symptom.

This is not an explanation. We are observing a stall, not throttling.

I'm going to take the unusual step of closing this PR, because I believe it is a shot in the dark, and I have no interest in merging hunches about bugs.

PR #525, opened 2026-05-21
Fix `communication_commitment` to hash value-only representation

JosephDenman:
There are two `communication_commitment` functions which compute `transient_commit` over the input/output pair as a bare `AlignedValue`. `<AlignedValue as FieldRepr>::field_repr` writes the alignment encoding before the value, so the result was a hash of `[alignment.., values..]` rather than of the values alone.

Everything else commits over the value-only representation. `Intent::add_call` builds the commitment that's stored on a ContractCall from `value_only_field_repr` while `StandardTransaction.add_calls` builds the commitment from the alignment encoding. I'm not _certain_ this is a bug but it looks like one.


tkerber (commented):
Could you add to `CHANGELOG_onchain_runtime.md`?

JosephDenman:
> Could you add to `CHANGELOG_onchain_runtime.md`?

Done

tkerber (approved):

PR #530, opened 2026-05-22
Enforce AlignedValue alignment in Key and StateValue deserialization

tkerber:
## Description

`AlignedValue` didn't check alignment on deserialization. That's *mostly* harmless, but we should fix anyway. To make this backwards compatible, we're checking at the point of use – in the onchain runtime. There's two of these: `Key` and `StateValue`. The former is ephemeral, and just in transcripts, so checking there is fine. The latter, we check in the state invariant, so that existing state is still accepted.

Note that formally this is *not* a data-format change, as old *state* will still be accepted.

## Sanity Checklist

This PR:

### DIFF
--- a/base-crypto/src/fab/encoding.rs
+++ b/base-crypto/src/fab/encoding.rs
@@ -535,6 +535,26 @@ impl AlignedValue {
         }
     }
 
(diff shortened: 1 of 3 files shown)


AlexShielded (approved):
LGTM

Issue #413, opened 2026-04-14
New bridge system transactions


tkerber:
## Describe the proposed feature:

Concrete planning for cNIGHT -> mNIGHT bridge transactions are taking shape, and will require slightly different transfers than currently. In particular, we need to support direct transfers (of invalid transactions *to* the bridge) into the treasury, and aggregate transfers of sub-threshold transactions, also to the treasury.

## User Need:

Need of the bridge subsystem to handle invalid transfers without loss of total funds.

## Expected Benefit:

More ability for the treasury to gather funds.

## Details:

tbd

tkerber:
Implementation work: #505

jacek-kurkowski-shielded:
LGTM, tests have been merged.

PR #534, opened 2026-05-25
test: integration tests for locked-pool unlock gestures

jacek-kurkowski-shielded:
## Description

Tests for: https://github.com/midnightntwrk/midnight-ledger/issues/413

## Sanity Checklist

This PR:

### DIFF
--- a/ledger-wasm/src/state.rs
+++ b/ledger-wasm/src/state.rs
@@ -22,7 +22,7 @@ use base_crypto::cost_model::{FixedPoint, NormalizedCost};
 use base_crypto::time::Timestamp;
 use coin_structure::coin::UserAddress;
 use js_sys::{Array, BigInt, Date, Function, Map, Set, Uint8Array};
-use ledger::structure::{ClaimKind, OutputInstructionUnshielded};
+use ledger::structure::{ClaimKind, INITIAL_PARAMETERS, OutputInstructionUnshielded};
 use onchain_runtime_wasm::from_value_ser;
 use onchain_runtime_wasm::state::ChargedState;
 use rand::Rng;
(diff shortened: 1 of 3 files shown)


ledger-wasm/src/state.rs
```diff
@@ -247,6 +247,53 @@ impl LedgerState {
...
+            INITIAL_PARAMETERS,
+            locked,
+            reserve,
+            0,
+        )
+        .map_err(|e| JsError::new(&format!("invalid genesis settings: {e:?}")))?;
+        Ok(LedgerState(state))
+    }
+
+    #[wasm_bindgen(js_name = "testingUnlockToTreasury")]
```
jsidorenko, line 275: do we really need those new testing_* methods here? As we already have the `apply_system_tx()` api method

ledger-wasm/src/state.rs
```diff
@@ -247,6 +247,53 @@ impl LedgerState {
...
+        let locked =
+            u128::try_from(locked_pool).map_err(|_| JsError::new("locked_pool is out of range"))?;
+        let reserve = u128::try_from(reserve_pool)
+            .map_err(|_| JsError::new("reserve_pool is out of range"))?;
+        let state = ledger::structure::LedgerState::<InMemoryDB>::with_genesis_settings(
+            network_id,
+            INITIAL_PARAMETERS,
+            locked,
+            reserve,
+            0,
```
jsidorenko: given it's a testing method, I think we should allow changing the last `treasury` param as well instead of forcing it to 0

ledger-wasm/src/state.rs
jacek-kurkowski-shielded: We don't want to expose `SystemTransactions` at the WASM level.
Using them isn't the responsibility of dApp developers.

ledger-wasm/src/state.rs
jacek-kurkowski-shielded: Well, after a careful 5 minut thinking I think you're right 😄 
Let me add it.

jsidorenko (approved):

PR #535, opened 2026-05-26
`zk-stdlib` v2

AlexShielded:
## Description

Update `zk-stdlib` to v2. Keeps v1 around in order to verify proofs generated against v1.

## Sanity Checklist

This PR:
- [ ] is primarily authored by AI, tests for `v1` -> `v2` verification authored by AI, as was the re-instantiation of `v1` logic


transient-crypto/.claude/settings.local.json
tkerber, line 1: :-1:, remove

transient-crypto/src/proofs.rs
```diff
@@ -620,6 +679,64 @@ impl VerifierKey {
...
+    }
+
+    /// Checks a proof against a statement using zk-stdlib v1.
+    pub fn verify_v1<F: Iterator<Item = Fr>>(
+        &self,
+        params: &ParamsVerifierV1,
+        proof: &Proof,
+        statement: F,
+    ) -> Result<(), VerifyingError> {
+        let vk = self.force_init_v1()?;
```
tkerber: I don't think this is the way to go for a few reasons:
- `force_init_v1`, unlike `force_init`, does _not_ cache the initialized key for reuse, meaning we need to pay to reinit the key on each verification. That's probably the case normally, but it's still not great.
- I'm not sure why we need to modify `transient-crypto` at all. Why not just pull in an old version of transient crypto in `onchain-state` / `ledger` and use the verification code from there for `ContractOperation::V3`?

tkerber (changes requested):

transient-crypto/src/proofs.rs
AlexShielded: Ah, that's a good point. I might have been a bit code blind while working directly within `transient-crypto`.

PR #564, opened 2026-06-05
`zk-stdlib` v2

AlexShielded:
## Description

This PR upgrades zk-stdlib to v2. This is a breaking change to `transient_crypto` so downstream ledger crates have to dynamically switch between the old version and the new version of crates. v1 is the default for proving. For verification we should be able to switch between v1 and v2 depending on the verifier key.

## Sanity Checklist

This PR:


zkir/tests/proofs_v2.rs
```diff
@@ -0,0 +1,208 @@
...
+            .prove(
+                &mut ChaCha20Rng::from_seed([42; 32]),
+                &TestParams,
+                pk,
+                &preimage,
+            )
+            .await
+            .unwrap();
+
+        // Wrong statement should fail
```
jacek-kurkowski-shielded: No shit AI 😄

AlexShielded: Well... it isn't wrong ;)

jacek-kurkowski-shielded (approved):
Looks good 👍

proof-server/src/versioned_ir.rs
```diff
@@ -85,14 +85,11 @@ pub(crate) async fn prove(
...
     resolver: &Resolver,
 ) -> Result<(Proof, Vec<Option<usize>>), String> {
-    if let Ok(_ir_v2) = zkir_v2::IrSource::load_from_tagged(Cursor::new(ir_source)) {
-        ppi.prove::<zkir_v2::IrSource>(OsRng, &*PUBLIC_PARAMS, resolver)
-            .await
-            .map_err(|e| e.to_string())
-    } else {
-        Err("Unsupported ZKIR version".into())
-    }
+    let proof = zkir_v2::ir_v1::v1_prove(&ppi, OsRng, &*PUBLIC_PARAMS, resolver)
```
tkerber, line 91: We want to switch here and also do v2 proving.

PR #565, opened 2026-06-05
ZKIR v3: extend `mul`, `neg`, and `inv` arithmetic to all supported types

miguel-ambrona:
## Description

Makes `mul` and `neg` polymorphic over all relevant IR types.

Introduces a new `inv` instruction for multiplicative inversion on field types.

## Sanity Checklist

This PR:


zkir-v3/src/ir_instructions/mul.rs
```diff
@@ -0,0 +1,129 @@
...
+        ))),
+    }
+}
+
+impl Mul for IrValue {
+    type Output = Self;
+
+    fn mul(self, rhs: Self) -> Self {
+        mul_offcircuit(&self, &rhs).unwrap()
+    }
```
iquerejeta, line 94: Can you avoid this unwrap by defining `type Output = Result<Self, Error>`?

zkir-v3/src/ir_instructions/neg.rs
```diff
@@ -0,0 +1,135 @@
...
+        ))),
+    }
+}
+
+impl Neg for IrValue {
+    type Output = Self;
+
+    fn neg(self) -> Self {
+        neg_offcircuit(&self).unwrap()
+    }
```
iquerejeta, line 107: Same as with mul - can we avoid unwrap?

iquerejeta (approved):
Looks good to me - minor comments on unwraps for off-circuit mul/neg.

zkir-v3/src/ir_instructions/neg.rs
miguel-ambrona: Nope

zkir-v3/src/ir_instructions/mul.rs
miguel-ambrona: I labeled the implementation of those traits as `cfg(test)` to make sure they are not used in the real execution.

PR #581, opened 2026-06-11
Prepare for rc.2 release.

tkerber:
## Description

Fixes dependency resolution issues caused by:
1. Hard pinning a specific version of `transient-crypto`
2. Not carrying the semver major `transient-crypto` bump into `coin-structure`

## Sanity Checklist

This PR:


Cargo.toml
```diff
@@ -64,3 +64,6 @@ panic = "abort"
 
 [workspace.metadata.spellcheck]
 config = ".spellcheck/cargo-spellcheck.toml"
+
+[patch.crates-io]
```
AlexShielded, line 68: We dont want to release with the patch, right?

AlexShielded (dismissed):
Small clarification

Cargo.toml
tkerber: No, we do. The patch is _required_, because we need the old `midnight-zkir` `serialize` dependency to unify with the local serialize. They _are_ both the same crate (`^1.0.0`), but one resolves via crates.io, while the other resolves as local.

That's _only_ a problem in the context of this repo, as elsewhere people will use the isolated tags or the crates.io releases, which will unify. In this repo, we need to patch the crates.io one to use the local one to unify instead.

AlexShielded (approved):

PR #583, opened 2026-06-15
Unified signing envelopes

tkerber:
## Description

Unify signing/commitment envelopes, and ensure they remain synced to their source structs.

Technically a format change, as the signing envelopes have changed.

## Sanity Checklist

This PR:


AlexShielded (approved):
Looks great!

PR #588, opened 2026-06-17
Another pass at `zk-stdlib` v2 upgrade

AlexShielded:
## Description

Another pass at the backwards compatible `zk-stdlib` v2 upgrade, incl. `proof-server` updates.

## Sanity Checklist

This PR:


proof-server/src/versioned_ir.rs
```diff
@@ -69,10 +69,19 @@ pub(crate) async fn prove(
     ir_source: &[u8],
     resolver: &Resolver,
 ) -> Result<(Proof, Vec<Option<usize>>), String> {
-    if let Ok(_ir_v2) = zkir_v2::IrSource::load_from_tagged(Cursor::new(ir_source)) {
-        ppi.prove::<zkir_v2::IrSource>(OsRng, &*PUBLIC_PARAMS, resolver)
-            .await
-            .map_err(|e| e.to_string())
+    if let Ok(ir_v2) = zkir_v2::IrSource::load_from_tagged(Cursor::new(ir_source)) {
+        match ir_v2.version {
+            zkir_v2::IrMinorVersion::V0 => {
```
tkerber: V1 already exists and uses zk-stdlib v1.

proof-server/src/versioned_ir.rs
```diff
@@ -69,10 +69,19 @@ pub(crate) async fn prove(
...
+                let proof = zkir_v2::ir_v1::v1_prove(&ppi, OsRng, &*PUBLIC_PARAMS, resolver)
+                    .await
+                    .map_err(|e| e.to_string())?;
+                Ok((proof, vec![]))
+            }
+            _ => ppi
```
tkerber: Explicit match here seems like a better option.

zkir/src/ir.rs
```diff
@@ -147,6 +147,13 @@ impl Zkir for IrSource {
         pk: ProverKey<Self>,
         preimage: &ProofPreimage,
     ) -> Result<(Proof, Vec<Fr>, Vec<Option<usize>>), ProvingError> {
+        if self.version == IrMinorVersion::V0 {
```
tkerber: Again, we already have a V1 using zk-stdlib v1.

zkir/Cargo.toml
tkerber: Whole point of this exercise was to eliminate this dependency.

zkir/src/ir.rs
```diff
@@ -178,61 +184,68 @@ impl Zkir for IrSource {
     }
 
     fn k(&self) -> u8 {
-        match self.to_v1() {
-            Ok(old_ir) => {
+        match self.version {
+            IrMinorVersion::V0 => {
```
tkerber: Ditto

zkir/src/ir.rs
```diff
@@ -147,6 +147,13 @@ impl Zkir for IrSource {
         pk: ProverKey<Self>,
         preimage: &ProofPreimage,
     ) -> Result<(Proof, Vec<Fr>, Vec<Option<usize>>), ProvingError> {
+        if self.version == IrMinorVersion::V0 {
+            anyhow::bail!(
+                "Zkir::prove is not supported for V0 circuits; \
+                 use ir_v1::v1_prove for the v1 proving pipeline"
```
tkerber: This is not what we agreed on, we agreed to implement the old transient-crypt's `Zkir` for the same `IrSource`. This does not preserve backwards compatibility.

tkerber (changes requested):

transient-crypto/src/proofs.rs
```diff
@@ -508,6 +508,17 @@ impl Serializable for VerifierKey {
 }
 
 impl VerifierKey {
+    /// Marks this verifier key as v1-only, preventing initialization with the
+    /// v2 `MidnightVK::read`. V1 VKs should only be verified through
+    /// `ir_v1::v1_verify`.
+    pub fn mark_v1_only(&self) {
```
tkerber: Why is this necessary? This seems quite annoying long-term baggage. Also, the implementation makes little sense to me – it's coercing uninitialized keys to invalid – why? Why should this be considered a safe operation?

transient-crypto/src/proofs.rs
```diff
@@ -528,10 +539,16 @@ impl VerifierKey {
             InnerVerifierKey::Uninitialized(data) => data.clone(),
         };
         let reader = &mut &data[..];
-        let vk = MidnightVK::read(reader, SerdeFormat::Processed)
-            .map_err(|_| anyhow::anyhow!("problem reading the verifier key"))?;
-        *mutex = InnerVerifierKey::Initialized(vk.clone(), data);
-        Ok(vk)
+        match MidnightVK::read(reader, SerdeFormat::Processed) {
```
tkerber: `match` seems like an odd choice here, although the `is_empty` check does indeed make sense. Still, I don't see the strict need here, I'd prefer this PR be limited to ZKIR.

transient-crypto/src/proofs.rs
```diff
@@ -761,17 +778,7 @@ impl ProofPreimage {
         let prover_key =
             Z::load_prover_key_from_tagged(io::Cursor::new(&proof_data.prover_key[..]))?;
         let (proof, pis, pi_skips) = ir.prove(rng, params, prover_key, self).await?;
-        debug!("proof created; verifying to make sure");
-        let k = verifier_key.force_init()?.k();
```
tkerber, line 765: :-1:, this is a very useful self-check that has saved us multiple times. Also unrelated to the purpose of this PR.

zkir/src/ir.rs
```diff
@@ -124,12 +123,27 @@ pub enum IrMinorVersion {
...
+/// (cannot be initialized with v2 `MidnightVK::read`).
+fn v1_vk_to_current(
+    old_vk: &transient_crypto_old::proofs::VerifierKey,
+) -> Result<VerifierKey, anyhow::Error> {
+    let mut buf = Vec::new();
+    serialize_old::Serializable::serialize(old_vk, &mut buf)?;
+    // Deserialize to extract the inner bytes (stripping the length prefix),
+    // then mark as v1-only so force_init() won't try to parse with v2 format.
+    let vk: VerifierKey = serialize::Deserializable::deserialize(&mut &buf[..], 0)?;
+    vk.mark_v1_only();
```
tkerber: So the purpose is to make it unusable? Then why do we need it in the first place?

zkir/src/ir.rs
```diff
@@ -124,12 +123,27 @@ pub enum IrMinorVersion {
     V0,
     #[default]
```
tkerber: Move `#[default]` marker?

zkir/src/ir.rs
```diff
@@ -150,89 +164,81 @@ impl Zkir for IrSource {
...
-            let vk: VerifierKey = serialize::Deserializable::deserialize(&mut &raw[..], 0)?;
-            Ok(vk)
-        } else {
-            use midnight_zk_stdlib::setup_vk;
-            let k = midnight_zk_stdlib::optimal_k(self) as u8;
-            Ok(VerifierKey::from(setup_vk(params.get_params(k).await?.as_ref(), self)))
+        match self.version {
+            IrMinorVersion::V0 | IrMinorVersion::V1 => {
+                let old_vk = crate::ir_v1::v1_keygen_vk(self, params).await?;
+                Ok(v1_vk_to_current(&old_vk)?)
```
tkerber: Just error out?

zkir/src/ir.rs
```diff
@@ -124,12 +123,27 @@ pub enum IrMinorVersion {
...
+    let mut buf = Vec::new();
+    serialize_old::Serializable::serialize(old_vk, &mut buf)?;
+    // Deserialize to extract the inner bytes (stripping the length prefix),
+    // then mark as v1-only so force_init() won't try to parse with v2 format.
+    let vk: VerifierKey = serialize::Deserializable::deserialize(&mut &buf[..], 0)?;
+    vk.mark_v1_only();
+    Ok(vk)
+}
+
 impl Zkir for IrSource {
```
tkerber, line 116: I was still expecting to see a `impl transient_crypto_old::Zkir for IrSource`.

tkerber (changes requested):

transient-crypto/src/proofs.rs
AlexShielded: Way ahead of you here. Just working on a change that peeks at the tags and gives us V1 verifier keys that way

zkir/src/ir.rs
AlexShielded: I thought we wanted the v1 pipeline to be the default?

tkerber: I don't see why.

zkir/src/ir_vm.rs
```diff
@@ -863,9 +863,9 @@ impl Relation for IrSource {
             .instructions
             .iter()
             .any(|op| matches!(op, I::PersistentHash { .. }));
-        let nr_pow2range_cols: u8 = match self.version {
-            IrMinorVersion::V0 => 1,
-            IrMinorVersion::V1 => 4,
+        let nr_pow2range_cols = match self.version {
+            IrMinorVersion::V0 | IrMinorVersion::V1 => 1,
```
tkerber: Wrong grouping!

zkir/src/ir_vm.rs
```diff
@@ -606,7 +606,7 @@ impl Relation for IrSource {
         for ins in self.instructions.iter() {
             match ins {
                 I::Assert { cond } => std.assert_non_zero(layouter, idx(&memory, *cond)?)?,
-                I::CondSelect { bit, a, b } if self.version == IrMinorVersion::V0 => {
+                I::CondSelect { bit, a, b } if self.version != IrMinorVersion::V2 => {
```
tkerber: This is rewriting the rules for `V1`! (Ditto for two diffs below)

zkir/src/lib.rs
```diff
@@ -16,18 +16,99 @@ extern crate tracing;
...
 mod ir;
-pub mod ir_v1;
+pub(crate) mod ir_v1;
 mod ir_vm;
 
 pub use ir::{Instruction, IrMinorVersion, IrSource, VersionedInnerPK};
 pub use ir_vm::Preprocessed;
 
+/// The tag used by v1 (old) verifier keys.
+const V1_VK_TAG: &str = "verifier-key[v6]";
```
tkerber: Better to use `transient_crypto_old::proofs::VerifierKey::tag()`?

zkir/src/lib.rs
tkerber: Insuffient; We explicitly need `impl<'a, R: Rng + CryptoRng + SplittableRng, S: transient_crypto_old::proofs::Resolver, P: transient_crypto_old::proofs::ParamsProvingProvider> transient_crypto_old::proofs::ProvingProvider for LocalProvingProvider<'a, R, S, P>`.

tkerber (changes requested):

.claude/settings.local.json
tkerber, line 1: :-1:

ledger/Cargo.toml
```diff
@@ -24,11 +24,13 @@ onchain-runtime = { version = "4.0.0", path = "../onchain-runtime", package = "m
 zswap = { version = "9.0.0", path = "../zswap", package = "midnight-zswap" }
 base-crypto = { version = "1.1.0", path = "../base-crypto", package = "midnight-base-crypto" }
 transient-crypto = { version = "3.0.0", path = "../transient-crypto", package = "midnight-transient-crypto" }
+transient-crypto-old = { package = "midnight-transient-crypto", version = "^2.0.0" }
+ff = "^0.13.0"
```
tkerber: Curious – why do we need `ff` directly? That seems like a smell.

zkir/src/ir.rs
```diff
@@ -59,7 +57,7 @@ impl VersionedInnerPK {
 /// A low-level IR allowing the prover to populate circuit witnesses.
 #[cfg_attr(feature = "proptest", derive(Arbitrary))]
 #[derive(Default, Clone, Debug, PartialEq, Serialize, Deserialize, Serializable)]
-#[tag = "ir-source[v2-generic]"]
+#[tag = "ir-source[v3]"]
```
tkerber: Negative. V3 is also already planned. Please stick to `v2-generic`, this is _intended_ to be modifiable.

tkerber (changes requested):
We're down to only a few nits now!

ledger/Cargo.toml
AlexShielded: We do some converting in ledger/src/structure.rs, but I can remove it

tkerber (dismissed):

AlexShielded:
Reviewed up to this point, LGTM. Cant submit a formal review as it's my PR.

AlexShielded (commented):
LGTM

tkerber (approved):
Approved in conjunction with Alex.

PR #648, opened 2026-07-17
ZKIR v3: Introduce `Bool` and `Byte` types

miguel-ambrona:
## Description

Introduce 2 new IR types and implement the natural instructions for them.

## Sanity Checklist

This PR:


zkir-v3/src/ir_vm.rs
```diff
@@ -758,6 +789,26 @@ impl Relation for IrSource {
...
+            if operands.is_empty() {
+                return Err(Error::Synthesis(
+                    "Boolean gate requires at least one input".into(),
+                ));
+            }
+            operands
+                .iter()
+                .map(|op| {
+                    let val = resolve_operand(std, layouter, memory, op)?;
+                    AssignedBit::try_from(val)
```
iquerejeta, line 807:
```suggestion
                    AssignedBit::try_from(val)
                        .map_err(|_| anyhow!("Boolean gate expects Bool inputs, found {t:?}"))
```

iquerejeta (approved):
LGTM - given that we implement AND, OR and XOR only for bits, sounds reasonable to add functions to convert to/from bits.

PR #661, opened 2026-07-24
[WIP] Verify Proof in ZKIR

iquerejeta:
## Description

Adds a verify_proof instruction to ZKIR-v3 that verifies an inner proof in-circuit via midnight-zk's verifier gadget.

## Sanity Checklist

This PR:

## Details

* VK is stored by hash. The instruction holds only `hash(vk)` (of the VK blob); the full VK is resolved out-of-band into `IrSource.verify_proof_vks`. This keeps the VK out of the canonical IR. keygen/prove check that the resolved blob hashes to `vk_hash` and fail fast otherwise.
* Proof is a witness. The inner proof lives in `ProofPreimage.proof_witnesses`, consumed positionally.
* Public inputs constrained in statement order (load-bearing). `verify_proof` exposes the inner accumulator mid-synthesis and records its offset from the live public-input counter; the offset only matches the statement vector if every PI is constrained in order.
* ZKIR forwards the opaque vk_blob, instance and proof to midnight-zk's byte entry points.
* Verification; a single verify runs the Plonk check and the inner proof's deferred pairing.


transient-crypto/src/proofs.rs
```diff
@@ -722,6 +722,11 @@ pub struct ProofPreimage {
     pub public_transcript_inputs: Vec<Fr>,
     /// A public statement vector encoding statement call results in the IR.
     pub public_transcript_outputs: Vec<Fr>,
+    /// Opaque, prover-supplied byte blobs consumed positionally by `VerifyProof`
+    /// instructions — one blob per `VerifyProof`, in instruction order. These
+    /// are the inner proofs being verified in-circuit; the proof system does not
+    /// interpret them (they are handed verbatim to the decider as a witness).
+    pub proof_witnesses: Vec<Vec<u8>>,
```
tkerber, line 876: On the one hand, having these be opaque is nice – on the other, I think we would eventually want a proving flow where the proof server / prover can unwind a nesting of these by itself.

I wonder if we could reasonably have this be something live `Vec<Either<Proof, ProofPreimage>>` instead.

I also think that while 'opaque' is helpful on the one hand, it leaves me with more questions than answers – is this the same encoding as 'regular' proofs, or is there a difference?

transient-crypto/Cargo.toml
```diff
@@ -27,10 +27,12 @@ storage-core = { version = "1.0.0", path = "../storage-core", package = "midnigh
 
 ff = "^0.13.1"
 group = "^0.13.0"
-midnight-curves = { version = "^0.3.1" }
-midnight-proofs = { version = "^0.8.1" }
-midnight-circuits = { version = "^7.2.2" }
-midnight-zk-stdlib = { version = "^2.3.3" }
+# Pointed at the `iquerejeta/verify_proof_stdlib` branch via `[patch.crates-io]`
+# in the workspace root. Versions match that branch's crates.
+midnight-curves = { version = "=0.3.0" }
```
tkerber: Note that I'm a massive hater of `=` versions. They break _so_ much stuff when you have them as transitive dependencies; I won't want to merge anything with these in full stop. They're also not necessary for patches to be applied.

transient-crypto/src/proofs.rs
tkerber, line 1: No changes to verification logic? :o 

Or is that hidden in the midnight-zk upgrade?

iquerejeta: It is hidden in the midnight-zk upgrade. When you call the `verify_proof` function, it tracks the position of the public inputs where the decider is, and the verification happens in the standard verification function.

transient-crypto/Cargo.toml
iquerejeta: No worries - I just got a bit of help of claude for fetching local repo, and the equal was added. Not planning to keep this in the final PR.

transient-crypto/src/proofs.rs
iquerejeta: I like the idea of having `Either<Proof, ProofPreimage>`. The issue I see with that is that we need to have the inner proving key be part of the outer proving key, and that's something that we do not want to support yet. IMO I would leave the design with `Either` when we can test the design with the proof server unwinding proofs. 

For inner proofs, the encoding is different. First we use a different hash function for the transcript, and secondly, we do not use the same wrapper as for proofs.

tkerber: > The issue I see with that is that we need to have the inner proving key be part of the outer proving key, and that's something that we do not want to support yet.

I don't think that's quite true – currently the provers (both proof server and wasm) assume an ad-hoc lookup from `key_location` -> key material, where key location is part of the proof preimage.

The inner proof using different settings sounds like something we will have to be very careful with here however...

transient-crypto/src/proofs.rs
tkerber: Makes sense. I assume this is a breaking change and will need version bumps etc.?

iquerejeta: yes, it will need version bumps indeed.

transient-crypto/src/proofs.rs
iquerejeta: For the moment the proof server cannot generate the inner proof. Are we OK moving forward with this design @tkerber ?

.tag-decompositions/proof-preimage
tkerber, line 1: Will need refactoring to shift proof-preimage to a new version tag. This is a disallowed edit.

ledger/src/construct.rs
```diff
@@ -559,6 +559,7 @@ impl<D: DB> ContractCallExt<D> for ProofPreimage {
         let binding_input = 0u8.into();
 
         let proof = ProofPreimage {
+            proof_witnesses: vec![],
```
tkerber, line 562: This *must* be a placeholder, otherwise this feature is unusable.

onchain-runtime-wasm/src/primitives.rs
```diff
@@ -254,6 +254,7 @@ pub fn proof_data_into_serialized_preimage(
     input.value_only_field_repr(&mut comm_comm_preimage);
     output.value_only_field_repr(&mut comm_comm_preimage);
     let preimage = transient_crypto::proofs::ProofPreimage {
+        proof_witnesses: vec![],
```
tkerber, line 257: Likewise, this *must* be a placeholder.

transient-crypto/src/proofs.rs
tkerber: Not quite. This isn't forward-compatible with an iteration which *could*. I think the bare minimum is to wrap things this part of the proof preimage in a unit-enum that could later be extended to add different inner-proof representations. e.g.

```rust
pub enum InnerProofWitness {
    Direct(Vec<u8>),
    // In the future maybe:
    // Preimage(ProofPreimage),
}
```

transient-crypto/src/proofs.rs
```diff
@@ -722,6 +872,15 @@ pub struct ProofPreimage {
...
     pub public_transcript_outputs: Vec<Fr>,
+    /// Opaque, prover-supplied byte blobs consumed positionally by `InnerProof`
+    /// instructions — one blob per *active* `InnerProof`, in instruction order.
+    /// A guarded-off `InnerProof` consumes nothing and binds the empty blob, so
+    /// this vector carries proofs only for the path actually taken, and a blob
+    /// left unconsumed is an error. These are the inner proofs verified
+    /// in-circuit by the `VerifyProof` instructions that consume them; the proof
+    /// system does not interpret them (they are handed verbatim to the decider
+    /// as a witness).
+    pub proof_witnesses: Vec<Vec<u8>>,
```
tkerber, line 883: Practical blockers:
- This mandates a proof-preimage tag bump
- This mandates a transient-crypto major version bump

zkir-precompiles/simple-merkle-tree/check.verifier.sha256
tkerber, line 1: Why did these change? This a strict failure condition if incidental – any proof system edits should *not* require old contracts to update.

zkir-v3/src/ir.rs
```diff
@@ -47,6 +47,12 @@ pub struct IrSource {
     pub do_communications_commitment: bool,
     /// The sequence of instructions to run in-circuit
     pub instructions: Arc<Vec<Instruction>>,
+    /// Full verifying keys for the circuit's `VerifyProof` instructions. 
+    /// Each entry is
+    /// [`serialize_vk`](crate::ir_instructions::decidable::serialize_vk)'s
+    /// output: the declared `DeciderKind`'s tag byte, then the `MidnightVK`.
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub verify_proof_vks: Vec<Vec<u8>>,
```
tkerber, line 55: This is backwards-compatibility-breaking for `Serializable`. That's fixable – we can branch on `version` to determine if this field exists or not, but that needs to be done manually. And, obviously, `version` *does* need to bump.

transient-crypto/src/proofs.rs
tkerber: Also, clarify please: Why does this need to be a new field? Conceptually it feels like this is could be part of the public transcript? I'm fine with it *being* a new field, but let's document the reason.

zkir-v3/src/ir.rs
tkerber: Also, is there a reason the VKs are here, and not embedded in `Instruction`?

zkir/src/ir_v1.rs
```diff
@@ -565,6 +565,7 @@ fn preimage_from_v1(
         transient_crypto::curve::Fr::from_le_bytes(&f.as_le_bytes()).expect("Fr round-trip")
     };
     transient_crypto::proofs::ProofPreimage {
+        proof_witnesses: vec![],
```
tkerber, line 568: :-1: – this will be tracking a *new* version of transient-crypto / midnight-zk. I'd rather not do that again for zkir v2, let's just target the existing old versions.

tkerber (changes requested):

zkir-precompiles/simple-merkle-tree/check.verifier.sha256
iquerejeta: The verifying key is changing (now we store the offsets of the accumulators in the public inputs, so that the verifier can access them)

zkir-v3/src/ir.rs
iquerejeta: We can bump the version here 👍

zkir/src/ir_v1.rs
iquerejeta: Alright - will take care of this with the version bump.

iquerejeta:
Closing in favour of #738

PR #662, opened 2026-07-28
proof-server: support GCP Confidential Space

julian-bcw:
## Description

This adds support for deploying the proof server in GCP's Confidential Space, using the `gcp_cs` feature. There's a new route:  `/attestation`, which returns a token (either OIDC or PKI) depending on the request header.

## Sanity Checklist

This PR:

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -78,7 +78,12 @@ jobs:
           # failures in `nix develop` command.
           set -o pipefail
           nix develop -c cargo config -Z unstable-options get net.git-fetch-with-cli
-          nix develop -c cargo test --workspace --all-features --release  --no-fail-fast -- --nocapture --test-threads=1 | tee test-output.log
+          {
+            test_status=0
+            nix develop -c cargo test --workspace --all-features --exclude midnight-proof-server --release --no-fail-fast -- --nocapture --test-threads=1 || test_status=$?
+            nix develop -c cargo test --package midnight-proof-server --features experimental --release --no-fail-fast -- --nocapture --test-threads=1 || test_status=$?
+            exit "$test_status"
+          } | tee test-output.log
 
       - name: Upload Artifact
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a
--- a/.github/workflows/test-coverage.yml
+++ b/.github/workflows/test-coverage.yml
@@ -65,7 +65,8 @@ jobs:
           export CARGO_INCREMENTAL=0
           export RUSTFLAGS='-Cinstrument-coverage'
           export LLVM_PROFILE_FILE='cargo-test-%p-%m.profraw'
-          nix develop -c cargo test --all --all-features --verbose -- --nocapture --test-threads=1
+          nix develop -c cargo test --workspace --all-features --exclude midnight-proof-server --verbose -- --nocapture --test-threads=1
+          nix develop -c cargo test --package midnight-proof-server --features experimental --verbose -- --nocapture --test-threads=1
 
       - name: Generate HTML report
         run: |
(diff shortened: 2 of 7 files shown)


Jalal-1:
Thanks Julian, the overall approach looks good. Four items to address before merge (detail shared separately):

1. **Reject unencrypted requests** — `gcp_cs` build currently falls back to plaintext on `/prove` and `/prove-tx` when encryption headers are absent ([endpoints.rs#L210-L219](https://github.com/midnightntwrk/midnight-ledger/blob/5b2124be11fdd5323285c2cf00e1997b8e7666c8/proof-server/src/endpoints.rs#L210-L219))
2. **Remove request-body logging** — decrypted payloads are logged at debug level; keep job ID / endpoint / size only ([`/prove`](https://github.com/midnightntwrk/midnight-ledger/blob/5b2124be11fdd5323285c2cf00e1997b8e7666c8/proof-server/src/endpoints.rs#L469-L472), [`/prove-tx`](https://github.com/midnightntwrk/midnight-ledger/blob/5b2124be11fdd5323285c2cf00e1997b8e7666c8/proof-server/src/endpoints.rs#L574-L577))
3. **Don't commit `200 OK` before job completion** — failures can't surface with the correct status once the streaming response starts ([endpoints.rs#L251-L274](https://github.com/midnightntwrk/midnight-ledger/blob/5b2124be11fdd5323285c2cf00e1997b8e7666c8/proof-server/src/endpoints.rs#L251-L274))
4. **Validate the attestation upstream response** — check status, add timeout and size limit before returning the token ([endpoints.rs#L364-L376](https://github.com/midnightntwrk/midnight-ledger/blob/5b2124be11fdd5323285c2cf00e1997b8e7666c8/proof-server/src/endpoints.rs#L364-L376))

Happy to re-review once these are in.

Jalal-1:
Thanks Julian — I’ve re-reviewed the latest commit. The four points from my original review are addressed: the `gcp_cs` build now rejects plaintext proving requests, decrypted request bodies are no longer logged, proof failures can return the correct HTTP status, and the attestation upstream response is bounded and validated.

Before we route this for formal Ledger/QA approval, could you please:

1. Rebase/update the branch against the current `ledger-8` — it is currently 50 commits behind.
2. Fix the `gcp_cs` integration-test path. The all-features coverage run enables `gcp_cs`, but existing tests still send plaintext and now receive `400` as expected.
3. Gate or adjust the new `proof-job-id` assertion, since that header is only emitted by the `gcp_cs` response path.
4. Rerun CI after the rebase. Some failures look runner-related, while the coverage failure above is deterministic.

Once the branch is current and CI is green, I’m happy to complete the review and get the required Ledger/QA reviewers involved.

julian-bcw:
@Jalal-1 thanks for your feedback, I've adjusted accordingly, and I think the only failures now are on the runners' side

Jalal-1:
Thanks Julian, appreciate you getting these changes in 🙏 I’ve had another look and the original points I raised have now been addressed. The remaining CI failures also look to be runner/dependency related from what I can see.

One thing to flag for the final technical review: `gcp_cs` is now excluded from the standard all-features Proof Server test run, so that specific path isn’t currently being exercised by CI.

@midnightntwrk/mn-codeowners-ledger @midnightntwrk/mn-qa — would you mind taking a final look from the Ledger/QA side and merging if you’re happy? Many thanks!

nstanford5:
@julian-bcw -- commits need verified signatures before this can be considered for merging.

syed-bcw:
> @julian-bcw -- commits need verified signatures before this can be considered for merging.

Taking this over since Julian has left. Successor PR: https://github.com/midnightntwrk/midnight-ledger/pull/765

The original commits were unsigned, so this work is being re-landed with verified signatures.

PR #666, opened 2026-07-31
Add array tests

jsidorenko:
## Sanity Checklist

This PR:


Cargo.lock
gilescope, line 3277: intentional?

jsidorenko: corresponds the reality

tkerber:
Superceded by #665 and regular forward-port.

PR #673, opened 2026-08-03
feat(ci): dedicated publish.yml with dual-scope npm publishing

### LINKED ISSUE #672: npm publishing: broken idempotence (wrong scope in existence checks) and no dedicated publish workflow
Three related gaps in the npm release path:

1. `prerelease.yml` builds `PKG` existence-check names with scope `@midnightntwrk`, but the published packages are `@midnight-ntwrk/*`. The "already published?" checks therefore never match, so re-runs after a partial failure attempt to republish and fail (`E409 Cannot publish over existing version` on GH Packages — see run 30651856895). The `force` delete path misses for the same reason.
2. Packages publish only under the `@midnight-ntwrk` scope. The org is migrating to `@midnightntwrk`; during the transition each version should publish under both scopes, on both registries (on GH Packages the new scope also attaches packages to the midnightntwrk org instead of the deprecated one).
3. npm publishing is inlined in `prerelease.yml` (three copy-pasted steps). npm trusted publishing validates the top-level workflow filename, so adopting it needs a dedicated, stable `publish.yml` that owns all npm publishes.

**Acceptance criteria:** re-running a partially-failed release skips already-published targets instead of erroring; each package version lands on GH Packages + npmjs under both scopes; a single `publish.yml` performs all npm publishes and is ready to register as the npm trusted publisher for all packages.

chrisferry:
Closes #672

## What

- **New `.github/workflows/publish.yml`** — the single npm-publish path. Dispatch inputs: `ref` (branch, tag or SHA), `packages` (`all`/`ledger`/`onchain-runtime`/`zkir`), `dry_run`. For each package it `nix build`s the tarball, reads name+version from the tarball itself, then **normalizes the tarball to each target scope** (`@midnight-ntwrk` legacy + `@midnightntwrk` migration target — the built scope changed on ledger-8 in f232211 and differs across release branches, so it is never trusted) and publishes idempotently to GH Packages + npmjs per scope. Gated by the `release` environment. Header documents the trusted-publisher registration (this one filename for all six package names) and the OIDC flip. Dry runs receive no registry credentials at all.
- **New `.github/scripts/rescope-npm-tarball.sh`** — repacks a tarball under a given scope; only `package.json` `name` changes (verified byte-identical file lists against the real `ledger-v8` tarball, both identity and cross-scope).
- **`prerelease.yml`** — inline npm publishing removed; the `@midnightntwrk`→`@midnight-ntwrk` scope typo in the three existence checks fixed (root cause of the `E409` re-run failures, run 30651856895); new **`publish-npm` job** (after build + coverage + integration + unit-tests) dispatches publish.yml pinned to `github.sha`, waits on the dispatched run, and propagates its conclusion; `release` requires it, so a GitHub release cannot complete with npm packages missing. npm publishing now happens after tests instead of before.
- **`npm-publish-idempotent.sh`** — registry version-conflicts (anchored `code E409`/`EPUBLISHCONFLICT`/"cannot publish over") are treated as the idempotent skip they imply.

## Unblocking 8.1.1

After merge: dispatch **Publish npm packages** with `ref=js/array-tests`. That branch builds `@midnight-ntwrk` tarballs (predates f232211); scope normalization publishes both scopes regardless. Already-published targets skip.

## Testing

- `actionlint` on both workflows (only pre-existing custom-runner/SC2035 notices remain), `shellcheck` clean on both scripts, `bash -n` on every run block
- `rescope-npm-tarball.sh` exercised against the real `@midnight-ntwrk/ledger-v8@8.1.0` tarball: identity and cross-scope repacks both yield 34/34 identical file entries and clean `npm publish --dry-run`
- Suggest a `dry_run=true` dispatch of publish.yml after merge as the live check

## Follow-ups (not this PR)

- Add required reviewers to the `release` environment (currently no protection rules) before registering publish.yml as the npm trusted publisher
- Register the six trusted publishers on npmjs, then apply the documented OIDC flip
- Confirm the `@midnightntwrk` org exists on npmjs and `MIDNIGHTCI_NPMJS_TOKEN` can publish to it (npmjs search shows zero packages under that scope today)

### DIFF
--- a/.github/scripts/npm-publish-idempotent.sh
+++ b/.github/scripts/npm-publish-idempotent.sh
@@ -87,4 +87,18 @@ publish_args=(--registry="$REG" "$SCOPE_REG_ARG" --tag "$TAG")
 if [ "$REG_HOST" = "registry.npmjs.org" ]; then
   publish_args+=(--access public)
 fi
-npm publish "$TGZ" "${publish_args[@]}"
+set +e
+OUTPUT=$(npm publish "$TGZ" "${publish_args[@]}" 2>&1)
+STATUS=$?
+set -e
+printf '%s\n' "$OUTPUT"
+if [ "$STATUS" -ne 0 ]; then
+  # A conflict means the version is already there (race, or a registry whose
+  # metadata endpoint was not readable by this token) - that is the idempotent
+  # success condition, not a failure.
+  if grep -qiE 'E409|EPUBLISHCONFLICT|cannot publish over' <<<"$OUTPUT"; then
+    echo "Version already exists on $REG_HOST (registry reported a conflict) - skipping"
+    exit 0
+  fi
+  exit "$STATUS"
+fi
--- a/.github/scripts/npm-publish-idempotent.sh
+++ b/.github/scripts/npm-publish-idempotent.sh
@@ -96,7 +96,9 @@ if [ "$STATUS" -ne 0 ]; then
   # A conflict means the version is already there (race, or a registry whose
   # metadata endpoint was not readable by this token) - that is the idempotent
   # success condition, not a failure.
-  if grep -qiE 'E409|EPUBLISHCONFLICT|cannot publish over' <<<"$OUTPUT"; then
+  # Anchored on 'code E409': npm's shasum/integrity notices are hex/base64
+  # that a bare 'E409' would occasionally match.
+  if grep -qiE 'code E409|EPUBLISHCONFLICT|cannot publish over' <<<"$OUTPUT"; then
     echo "Version already exists on $REG_HOST (registry reported a conflict) - skipping"
     exit 0
   fi
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
     inputs:
       force:
-        description: "Real runs only: overwrite an existing GH Packages version (delete-then-republish). Public npm cannot be overwritten and will error. Not needed for a dry run."
+        description: "Force a rebuild even when the version is already published on both registries. npm publishing itself lives in publish.yml (idempotent; existing versions are skipped)."
         type: boolean
         default: false
       dry_run:
@@ -27,6 +27,8 @@ jobs:
     environment: ${{ github.event.inputs.dry_run == 'true' && '' || 'release' }}
     permissions:
       contents: write
+      # Dispatch publish.yml (npm publishing) after tags are pushed.
+      actions: write
     outputs:
       tag_name: ${{ steps.ids.outputs.tag_name }}
       version: ${{ steps.ids.outputs.version }}
@@ -96,8 +98,6 @@ jobs:
       - name: Build and Publish Ledger
         env:
           NODE_AUTH_TOKEN: ${{ secrets.MIDNIGHTCI_PACKAGES_WRITE }}
-          NPM_TOKEN: ${{ secrets.MIDNIGHTCI_NPMJS_TOKEN }}
-          GH_TOKEN: ${{ secrets.MIDNIGHTCI_PACKAGES_WRITE }}
           FORCE: ${{ github.event.inputs.force }}
         run: |
           git config --global user.name 'MidnightCI'
@@ -113,19 +113,10 @@ jobs:
             ${SHORT_VERSION}.*) TAG_NAME="ledger-${VERSION}" ;;
             *)                  TAG_NAME="ledger-${SHORT_VERSION}.${VERSION}" ;;
           esac
-          PKG="@midnightntwrk/ledger-v${SHORT_VERSION}"
+          PKG="@midnight-ntwrk/ledger-v${SHORT_VERSION}"
           SCOPE="${PKG%%/*}"
           GH_REG="https://npm.pkg.github.com/midnightntwrk"
           NPMJS_REG="https://registry.npmjs.org/"
-
-          # Authenticate npmjs (GH Packages auth is configured by setup-node).
-          if [ -z "${NPM_TOKEN:-}" ]; then
-            echo "::error::MIDNIGHTCI_NPMJS_TOKEN is not set - cannot authenticate to npmjs"; exit 1
-          fi
-          # setup-node points npm at $NPM_CONFIG_USERCONFIG, so write the token
-          # there literally (npm's ${VAR} expansion yields empty -> ENEEDAUTH).
-          # Leading newline guards a setup-node npmrc with no trailing newline.
-          printf '\n//registry.npmjs.org/:_authToken=%s\n' "${NPM_TOKEN}" >> "${NPM_CONFIG_USERCONFIG:-$HOME/.npmrc}"
 
           # Build unless the version already exists on both registries (FORCE or
           # dry run always builds). The --<scope>:registry override is required:
@@ -145,13 +136,7 @@ jobs:
             # Copy out before the next nix build overwrites result/lib.
             mkdir -p "${GITHUB_WORKSPACE}/release-artifacts"
             cp "$TGZ" "${GITHUB_WORKSPACE}/release-artifacts/"
-            # Publish is idempotent per-registry: skips where the version already exists.
-            "${GITHUB_WORKSPACE}/.github/scripts/run-unless-dry.sh" \
-              "${GITHUB_WORKSPACE}/.github/scripts/npm-publish-idempotent.sh" \
-              "$TGZ" "$GH_REG" "$PKG" "${{ env.LEDGER_WASM_VERSION }}"
-            "${GITHUB_WORKSPACE}/.github/scripts/run-unless-dry.sh" \
-              "${GITHUB_WORKSPACE}/.github/scripts/npm-publish-idempotent.sh" \
-              "$TGZ" "$NPMJS_REG" "$PKG" "${{ env.LEDGER_WASM_VERSION }}"
+            # npm publishing happens in publish.yml, dispatched below.
             cd ../..
           fi
 
@@ -167,8 +152,6 @@ jobs:
       - name: Build and Publish Onchain Runtime
         env:
           NODE_AUTH_TOKEN: ${{ secrets.MIDNIGHTCI_PACKAGES_WRITE }}
-          NPM_TOKEN: ${{ secrets.MIDNIGHTCI_NPMJS_TOKEN }}
-          GH_TOKEN: ${{ secrets.MIDNIGHTCI_PACKAGES_WRITE }}
           FORCE: ${{ github.event.inputs.force }}
         run: |
           git config --global user.name 'MidnightCI'
@@ -177,19 +160,10 @@ jobs:
           VERSION="${ONCHAIN_RUNTIME_WASM_VERSION}"
           SHORT_VERSION=$(echo "$VERSION" | awk -F. '{print $1}')
           TAG_NAME="onchain-runtime-${VERSION}"
-          PKG="@midnightntwrk/onchain-runtime-v${SHORT_VERSION}"
+          PKG="@midnight-ntwrk/onchain-runtime-v${SHORT_VERSION}"
           SCOPE="${PKG%%/*}"
           GH_REG="https://npm.pkg.github.com/midnightntwrk"
           NPMJS_REG="https://registry.npmjs.org/"
-
-          # Authenticate npmjs (GH Packages auth is configured by setup-node).
-          if [ -z "${NPM_TOKEN:-}" ]; then
-            echo "::error::MIDNIGHTCI_NPMJS_TOKEN is not set - cannot authenticate to npmjs"; exit 1
-          fi
-          # setup-node points npm at $NPM_CONFIG_USERCONFIG, so write the token
-          # there literally (npm's ${VAR} expansion yields empty -> ENEEDAUTH).
-          # Leading newline guards a setup-node npmrc with no trailing newline.
-          printf '\n//registry.npmjs.org/:_authToken=%s\n' "${NPM_TOKEN}" >> "${NPM_CONFIG_USERCONFIG:-$HOME/.npmrc}"
 
           # Build unless the version already exists on both registries (FORCE or
           # dry run always builds). The --<scope>:registry override is required:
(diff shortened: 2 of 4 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,153 @@
...
+            cp result/lib/*.tgz "$ATTR_DIR/"
+            TGZ=$(find "$ATTR_DIR" -name '*.tgz' ! -name '*-rescoped.tgz' | head -1)
+            if [ -z "$TGZ" ]; then
+              echo "::error::nix build .#${ATTR} produced no npm tarball in result/lib"; exit 1
+            fi
+            PKG=$(tar -xzOf "$TGZ" package/package.json | jq -r '.name')
+            VER=$(tar -xzOf "$TGZ" package/package.json | jq -r '.version')
+            echo "::group::${PKG}@${VER} (${ATTR})"
+
+            RESCOPED=$("${GITHUB_WORKSPACE}/.github/scripts/rescope-npm-tarball.sh" "$TGZ" "$NEW_SCOPE" "$ATTR_DIR")
```
scottbuckel: [P1] The dual-scope publish is currently a no-op. The tarballs produced by these Nix attrs already use `@midnightntwrk`: `bagel.nix` defaults `scope ? "@midnightntwrk"`, and the `ledger-wasm`, `onchain-runtime-wasm`, and `zkir-wasm` definitions do not override it. Consequently `PKG` and `NEW_PKG` are identical here, the rescope step changes nothing, and the four targets below publish `@midnightntwrk` twice while never publishing `@midnight-ntwrk`. This also means the documented 8.1.1 recovery will not fill the missing old-scope npmjs target. Please either rescope the built tarball to the old scope or make the canonical build use the old scope, and test with the tarball produced by these exact Nix attrs.

.github/workflows/prerelease.yml
```diff
@@ -291,6 +242,19 @@ jobs:
             "${GITHUB_WORKSPACE}/.github/scripts/run-unless-dry.sh" git push origin "${TAG_NAME}"
           fi
 
+      - name: Dispatch npm publish
+        if: ${{ github.event.inputs.dry_run != 'true' }}
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          BUILD_REF: ${{ github.ref_name }}
```
scottbuckel: [P1] Please pass an immutable commit here (`github.sha`, or the tag that was just pushed) rather than `github.ref_name`. The dispatched publish job can sit behind the global concurrency group and a second `release` environment approval; if the branch advances before checkout, it will build and publish a different commit from the one this release workflow checked, tagged, and tested.

.github/workflows/prerelease.yml
```diff
@@ -291,6 +242,19 @@ jobs:
...
+        run: |
+          # npm publishing lives in publish.yml (the npm trusted-publisher
+          # workflow); its definition always runs from the default branch,
+          # building/publishing the ref this release was cut from.
+          DEFAULT_BRANCH=$(gh repo view "$GITHUB_REPOSITORY" --json defaultBranchRef --jq .defaultBranchRef.name)
+          gh workflow run publish.yml --repo "$GITHUB_REPOSITORY" \
```
scottbuckel: [P1] `gh workflow run` returns after creating the dispatch; it does not wait for the child workflow to start or succeed. The `build` job therefore goes green immediately and all downstream release jobs can finish while npm publishing is still queued, waiting for its own environment approval, or has failed. That can produce a successful/announced GitHub release with missing npm packages. Please make the parent wait for the exact dispatched run and propagate its conclusion (or otherwise make release completion depend on npm publication).

.github/scripts/npm-publish-idempotent.sh
```diff
@@ -87,4 +87,18 @@ publish_args=(--registry="$REG" "$SCOPE_REG_ARG" --tag "$TAG")
...
+set +e
+OUTPUT=$(npm publish "$TGZ" "${publish_args[@]}" 2>&1)
+STATUS=$?
+set -e
+printf '%s\n' "$OUTPUT"
+if [ "$STATUS" -ne 0 ]; then
+  # A conflict means the version is already there (race, or a registry whose
+  # metadata endpoint was not readable by this token) - that is the idempotent
+  # success condition, not a failure.
+  if grep -qiE 'E409|EPUBLISHCONFLICT|cannot publish over' <<<"$OUTPUT"; then
```
scottbuckel: `npm publish` writes `npm notice shasum: <40 hex chars>` (and the base64 integrity string) to stderr, which lands in `$OUTPUT` via `2>&1`. With `-i`, a bare `E409` pattern matches any shasum containing `e409` (~1 in 2000 tarballs), so an unrelated publish failure would be misreported as "already exists — skipping" and the pipeline would go green without publishing. Anchoring on `code E409` closes it:

```suggestion
  if grep -qiE 'code E409|EPUBLISHCONFLICT|cannot publish over' <<<"$OUTPUT"; then
```

.github/workflows/prerelease.yml
scottbuckel: `github.ref_name` is the branch, so publish.yml rebuilds from the branch HEAD *at publish time* — not the commit this release was built and tagged from. If the branch moves in between (or the publish run waits on environment approval once reviewers are configured), the published tarball comes from a different commit than the tag. `github.sha` pins it exactly, and publish.yml's `ref` input already accepts a SHA:

```suggestion
          BUILD_REF: ${{ github.sha }}
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,153 @@
...
+          NPMJS_REG="https://registry.npmjs.org/"
+          NEW_SCOPE="@midnightntwrk"
+
+          # npmjs auth (GH Packages auth is configured by setup-node). The
+          # token is written literally: npm's ${VAR} expansion yields empty
+          # -> ENEEDAUTH. Under trusted publishing this block is deleted.
+          if [ -z "${NPM_TOKEN:-}" ]; then
+            echo "::error::MIDNIGHTCI_NPMJS_TOKEN is not set - cannot authenticate to npmjs"; exit 1
+          fi
+          printf '\n//registry.npmjs.org/:_authToken=%s\n' "${NPM_TOKEN}" >> "${NPM_CONFIG_USERCONFIG:-$HOME/.npmrc}"
```
scottbuckel: Dry runs skip the `release` environment gate, but this step still receives `NPM_TOKEN` + `MIDNIGHTCI_PACKAGES_WRITE`, writes the npm token into `.npmrc`, and runs `nix build` on an arbitrary user-supplied `ref` — so anyone with write access can dry-run a crafted branch and exfiltrate both tokens with no approval. Inherited from prerelease.yml, so not a regression, but this new file is the right place to fix it: make the auth block conditional, e.g.

```bash
if [ "${DRY_RUN}" != "true" ]; then
  if [ -z "${NPM_TOKEN:-}" ]; then
    echo "::error::MIDNIGHTCI_NPMJS_TOKEN is not set - cannot authenticate to npmjs"; exit 1
  fi
  printf '\n//registry.npmjs.org/:_authToken=%s\n' "${NPM_TOKEN}" >> "${NPM_CONFIG_USERCONFIG:-$HOME/.npmrc}"
fi
```

(and ideally drop the two secrets from the step `env` on dry runs). Bonus: if `MIDNIGHTCI_NPMJS_TOKEN` ever moves to an environment-scoped secret, dry runs currently hard-fail on the not-set check — this fixes that too.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,153 @@
...
+  cancel-in-progress: false
+
+jobs:
+  publish:
+    # Standard hosted runner (free on public repos): the nix build is mostly
+    # restored from the cache shared with prerelease.yml, so no large runner.
+    runs-on: ubuntu-24.04
+    # Same approval gate as the release workflow: real publishes require the
+    # protected `release` environment; dry runs need no approval.
+    environment: ${{ github.event.inputs.dry_run == 'true' && '' || 'release' }}
```
scottbuckel: FYI: the `release` environment currently has **no protection rules** (checked via `gh api repos/midnightntwrk/midnight-ledger/environments/release` — empty `protection_rules`), so "same approval gate as the release workflow" is vacuously true today: both gates are open. Worth adding required reviewers before registering this workflow as the npm trusted publisher, since that registration cites `environment: release` as a security boundary. Once reviewers exist, note a release becomes one dispatch but two approvals (prerelease, then the chained publish run).

scottbuckel (commented):
Reviewed — the architecture is right, the scope-typo root cause matches the diff, and reading name/version from the tarball kills that whole bug class. Four inline comments below: two one-liners I'd fix before merge (anchored E409 grep, pin the dispatch to `github.sha`), two that can be fast-follows (dry-run secret exposure, `release` environment currently has no protection rules).

Nits, take or leave: `gh workflow run` already targets the default branch when `--ref` is omitted, so the `gh repo view … defaultBranchRef` lookup can be dropped; `find … | head -1` silently picks one tarball if a derivation ever emits several (count-and-error is safer); before the 8.1.1 unblock dispatch, confirm the `@midnightntwrk` scope exists on npmjs and `MIDNIGHTCI_NPMJS_TOKEN` can publish to it, otherwise the run dies at target 3 of 4 (re-run safe, but noisy).

_Review drafted with AI assistance (Claude Code)._

.github/workflows/publish.yml
chrisferry: Confirmed — you're right, and it explains an apparent contradiction: the Jul 31 run built `@midnight-ntwrk` tarballs because `js/array-tests` predates f232211, while ledger-8 now builds `@midnightntwrk`. Fixed in 4800f3f by not trusting the tarball's scope at all: every tarball is normalized to each target scope via the rescope script (identity repack when it already matches), so both scopes publish regardless of which branch/epoch the build came from. Tested both directions against the real ledger-v8 tarball.

.github/workflows/prerelease.yml
chrisferry: Fixed in 4800f3f — the dispatch now passes `github.sha`.

.github/workflows/prerelease.yml
chrisferry: Fixed in 4800f3f — dispatch moved out of `build` into a dedicated `publish-npm` job that dispatches pinned to `github.sha`, polls for the created run, then `gh run watch --exit-status` propagates its conclusion. `release` now needs `publish-npm` (skipped-aware for dry runs), so a GitHub release can't complete with npm packages missing. Side benefit: npm now publishes after coverage/integration/unit-tests instead of before them.

.github/scripts/npm-publish-idempotent.sh
chrisferry: Good catch — applied your suggestion verbatim in 4800f3f (`code E409` anchor), with a comment explaining why the anchor matters.

.github/workflows/publish.yml
chrisferry: Fixed in 4800f3f: the npmrc auth write is gated on a real run, and both secrets are withheld from the step env entirely on dry runs (`dry_run == 'true' && '' || secrets...`), same for the setup-node token. So a dry-run of a crafted ref has nothing to exfiltrate. Your bonus case (environment-scoped secret + dry run) works too since the not-set check only runs on real runs.

.github/workflows/publish.yml
chrisferry: Agreed — flagged to Chris to add required reviewers on the `release` environment before the trusted-publisher registration cites it. That's a repo-settings change, kept out of this PR.

.github/workflows/prerelease.yml
chrisferry: Fixed in 4800f3f together with the P1 — `github.sha` via the new publish-npm job.

chrisferry:
Pushed 4800f3f addressing the review — all four substantive findings were correct and are fixed (details on each thread). On the nits: the `defaultBranchRef` lookup is gone (moot in the new publish-npm job — `gh workflow run` without `--ref` targets the default branch); `find | head -1` replaced with a count-exactly-one guard that errors listing what it found; and on the pre-flight — npmjs search currently shows **zero** packages under `@midnightntwrk` and the org page isn't verifiable anonymously, so before the 8.1.1 unblock dispatch someone should confirm the npmjs org slug exists and `MIDNIGHTCI_NPMJS_TOKEN` has publish rights on it (a failed target is re-run-safe either way). The `release` environment protection is flagged as a follow-up for repo settings.

_Fixes authored with AI assistance (Claude Code)._

.github/workflows/publish.yml
```diff
@@ -0,0 +1,161 @@
...
+          primary-key: nix-${{ runner.os }}-${{ hashFiles('**/*.nix', '**/flake.lock') }}
+          restore-prefixes-first-match: nix-${{ runner.os }}-
+          purge: true
+          purge-prefixes: nix-${{ runner.os }}-
+          purge-created: 0
+
+      - name: Build and publish
+        env:
+          # Dry runs never publish; give them no credentials at all.
+          NODE_AUTH_TOKEN: ${{ github.event.inputs.dry_run == 'true' && '' || secrets.MIDNIGHTCI_PACKAGES_WRITE }}
```
scottbuckel: [P1] These expressions still expose both credentials on dry runs. In GitHub expression semantics, the empty string is falsy, so when dry_run=true the expression evaluates as true AND empty, then falls through the OR arm to the secret. In other words this is the secret for both true and false; the same issue exists on the setup-node token above. The environment expression at line 64 has the same shape and likewise always resolves to release. Please reverse the condition so the secret/release value is the truthy arm and empty/null is the final fallback. GitHub documents the key rule here: empty strings are falsy.

.github/workflows/prerelease.yml
```diff
@@ -379,6 +328,41 @@ jobs:
...
+        run: |
+          MARK=$(date -u +%Y-%m-%dT%H:%M:%SZ)
+          gh workflow run publish.yml --repo "$GITHUB_REPOSITORY" \
+            -f "ref=$BUILD_SHA" -f packages=all
+          # The dispatch API returns no run id; poll for the run it created.
+          RUN_ID=""
+          for _ in $(seq 1 30); do
+            sleep 5
+            RUN_ID=$(gh run list --repo "$GITHUB_REPOSITORY" --workflow publish.yml \
+              --event workflow_dispatch --created ">=$MARK" --limit 1 \
```
scottbuckel: [P1] The timestamp does not identify the run created by this dispatch. If another manual publish or prerelease dispatch occurs after MARK and before this poll, --limit 1 selects that newer run; this parent can then release based on a different commit while its own publish is still queued or later fails. The workflow concurrency group does not remove queued/waiting runs from this query. Please pass a unique correlation value such as the parent run id as an input, include it in publish.yml run-name, and match that exact displayTitle (or consume the run URL/id emitted by gh workflow run when available).

.github/workflows/prerelease.yml
```diff
@@ -419,7 +403,10 @@ jobs:
           nix develop .#ci -c .github/scripts/cargo-publish-crates.sh "$VERSION"
 
   release:
-    needs: [ build, coverage, integration, unit-tests ]
+    # publish-npm is skipped on dry runs; `always() &&` keeps release eligible
+    # then, while real runs still hard-require npm publication success.
+    if: ${{ always() && needs.build.result == 'success' && needs.coverage.result == 'success' && needs.integration.result == 'success' && needs.unit-tests.result == 'success' && (needs.publish-npm.result == 'success' || (github.event.inputs.dry_run == 'true' && needs.publish-npm.result == 'skipped')) }}
```
scottbuckel: [P1] Please use !cancelled() here instead of always(). GitHub re-evaluates job conditions during cancellation, and always() remains true; with the explicit needs checks also true, this irreversible release job can continue creating/uploading the GitHub release and announcing it after an operator cancels the workflow. GitHub explicitly warns against always() for tasks that can suffer a critical failure. !cancelled() is still a status function, so it bypasses the implicit success check needed for the dry-run skipped dependency while honoring cancellation.

.github/workflows/publish.yml
gilescope: inverted the if so the default is safe.

gilescope (dismissed):

gilescope (approved):

PR #684, opened 2026-08-10
Improve dust circuit and validation

tkerber:
## Description

Tightens Dust circuit.

## Sanity Checklist

This PR:


jsidorenko (approved):

spec/dust.md
```diff
@@ -323,33 +323,30 @@ struct DustGenerationInfo {
...
 components:
 - A mapping from Night addresses to Dust addresses, to provide the Dust owner value for new Night outputs.
 - A sequential Merkle tree of `DustGenerationInfo`, which can be directly used
   to assert the presence of a specific Dust generation info in ZK proofs.
-- A corresponding set of the `DustGenerationUniquenessInfo`s, to prevent
-  collisions in these.
-- A mapping from Night UTXOs to their position in the Merkle tree.
+- A mapping from each generation's initial nonce to its position in the Merkle
+  tree. This is append-only, and doubles as the uniqueness check preventing two
+  generations from sharing an initial nonce.
```
miguel-ambrona, line 343: Can you please rewrite this? It is hard to follow.

spec/dust.md
```diff
@@ -507,6 +505,15 @@ fn dust_spend_valid(
     assert!(commitment_root = commitment_merkle_tree.root());
     assert!(generation_merkle_tree.contains(gen));
     assert!(genration_root = generation_merkle_tree.root());
+    // Bind both the spent UTXO's nonce and the generation info to a shared
+    // initial nonce. Without this the generation info is unconstrained: any
+    // entry in the generation tree could be paired with the spent UTXO, letting
+    // a spender substitute a higher-value backing Night to inflate the computed
+    // updated value. The first UTXO in a chain (seq 0) is bound to the public
+    // key, all subsequent ones to the secret key (see the nonce discussion above).
```
miguel-ambrona, line 516: Useful for the review, but I would remove it from the spec.

All assertions are necessary, do we wanna add a similar comment to the rest of assertions?

spec/dust.md
```diff
@@ -517,7 +524,7 @@ fn dust_spend_valid(
     let v_pre = updated_value(inp, gen, tnow, params);
```
miguel-ambrona, line 527:
```suggestion
    let v_pre = updated_value(dust, gen, tnow, params);
```

spec/dust.md
miguel-ambrona, line 799: Should `parent` be `parent_intent`? Where is `parent` defined otherwise?

miguel-ambrona, line 322: Is this `owner` field ever read? If not, could we remove it from `DustGenerationInfo`?

miguel-ambrona, line 320: I understand this struct exists so that the timestamp can be updated when the associated night is spent.

If that is the only purpose, then `value` could be part of the `DustOutput` instead and live with it forever.
`owner` is not necessary here, it can be removed.

The only fields that are really needed are the dust identifier and the `dtime`.
Using the `nonce` as the identifier seems fine given the uniqueness checks, but `nonce` is not a great name for it. Can we call it `dust_id` or similar?

miguel-ambrona (commented):
Leaving a few comments, do not consider this to be a full review.

kapke (approved):

tkerber:
@miguel-ambrona, I'm considering your comments non-blocking to get this merged. That's annoying, and we should revisit them, but unfortunately there's now pressure to stabilise ledger 9, and none of these seem strong reasons to push back.

Issue #688, opened 2026-08-11
Add `FromBytes(64)` for `Curve25519Scalar`


iquerejeta:
`FromBytes(64)` needed for implementing `ed25519`. We are adding `ToBytes(64)` just due to a language design decision, to be able to round trip (even if the most significant bytes will be zero).

PR #701, opened 2026-08-13
Add FromBytes(64) for Curve25519Scalar

AlexShielded:
## Description

Closes https://github.com/midnightntwrk/midnight-ledger/issues/688

## Sanity Checklist

This PR:


zkir-v3/src/ir_instructions/assign.rs
```diff
@@ -113,5 +113,11 @@ pub fn assign_incircuit(
             .scalar_field_chip()
             .assign_many(layouter, &convert_values::<k256::Fq>(values)?)
             .map(|xs| xs.into_iter().map(CircuitValue::Secp256k1Scalar).collect()),
+
+        IrType::Curve25519Scalar => std_lib
+            .curve25519()
+            .scalar_field_chip()
+            .assign_many(layouter, &convert_values::<curve25519::Scalar>(values)?)
+            .map(|xs| xs.into_iter().map(CircuitValue::Curve25519Scalar).collect()),
```
iquerejeta, line 121: Don't we have this already merged in ledger-9? Would make sense to rebase these branches with the latest tip of ledger-9. @miguel-ambrona, could you bump your chain of PRs?

zkir-v3/src/ir_instructions/from_bytes.rs
```diff
@@ -0,0 +1,222 @@
...
+    ir_instructions::F,
+    ir_types::{CircuitValue, IrType, IrValue},
+};
+
+/// Builds (off-circuit) a value of the given type from its byte representation.
+/// Supported for types (with their expected number of bytes):
+///  - Native           (32 bytes)
+///  - Secp256k1Base    (32 bytes)
+///  - Secp256k1Scalar  (32 bytes)
+///  - Curve25519Scalar (64 bytes)
```
iquerejeta, line 35: `Curve25519Scalar` also needs a function to/from 32 bytes (that's the actual size of the scalar field).

zkir-v3/src/ir_instructions/to_bytes.rs
```diff
@@ -0,0 +1,157 @@
...
+    ir_instructions::F,
+    ir_types::{CircuitValue, IrValue},
+};
+
+/// Converts (off-circuit) the given value into its byte representation.
+/// Supported on types (with their number of bytes):
+///  - Native           (32 bytes)
+///  - Secp256k1Base    (32 bytes)
+///  - Secp256k1Scalar  (32 bytes)
+///  - Curve25519Scalar (64 bytes)
```
iquerejeta, line 32: Same as the "from" comment. We'd need converting to 32 bytes. In this case we can probably not support both.

iquerejeta (commented):
I've been thinking this through, and I don't really like this design. A Curve25519 scalar is 32 bytes. The reason why we want to have `from_bytes64` is because we generate a scalar from the output of SHA512, but reviewing this PR, I realised it doesn't make sense to have to/from bytes64. 

The roundtrip argument doesn't really hold, because `to_bytes64(from_bytes64(foo)) != foo`, given that from bytes does a modular reduction. However, that counterargument also holds for bytes arrays of size 32. 

My suggestion is the following - fields have a generic `from_bytes(bytes)` function, that allows you to call it with any sized list of bytes (fixed at compile time). Then, each field has `to_bytes` which constructs an output of fixed size. Roundtrip holds if bytes represent an integer smaller than the field size: `to_bytes(from_bytes(foo)) == foo.resize(32, 0)`. 

Would be good to have @kmillikin 's input here.

zkir-v3/src/ir_instructions/assign.rs
miguel-ambrona: What do you want me to do? I thought we were going to close those PRs since a new repo has been opened for ZKIR.

PR #708, opened 2026-08-18
feat(storage-core): attach opaque metadata to GC roots to support later pruning decisions

Klapeyron:
## Description

Adds GC-root metadata to `storage-core` (behind `gc-v1`): a persisted root can be tagged with an opaque value (e.g. a block hash), and later released by providing only that value — no caller-side index from tags back to root hashes.

**Motivation.** `persist()` marks a root but records nothing about what it was persisted *for*. A pruning job (e.g. a node reclaiming ledger states once their blocks fall out of the state-pruning window) would otherwise have to maintain its own durable `tag → root` index in a separate database, with the cross-database crash-consistency reasoning that entails (independent WALs, remove-then-act ordering, retry logic, clamping against replayed decrements). Storing the tag in the same database as the root count it guards reduces the caller's lifecycle to three calls, each using only data in hand at that moment:

```rust
sp.persist();                                  // execution time (tag not yet known)
sp.set_root_metadata(block_hash);              // finality time (tag known)
backend.unpersist_by_metadata(&pruned_hashes); // pruning time (one lock, one call)
```

**API** (all `gc-v1`-gated):

- `Sp::persist_with_metadata(tag)` — persist and tag in one step, when the tag is known at persist time.
- `Sp::set_root_metadata(tag)` / `StorageBackend::set_root_metadata` — tag an already-persisted root without changing its persist count.
- `StorageBackend::unpersist_by_metadata(&[tags]) -> usize` — release every root tagged with one of the given values, decrementing each matching root once. Zero-clamped: roots at count 0 are skipped, so submitting values that exceed the actually held counts is safe. The whole batch runs under a single backend borrow with one scan of the tag table.
- `StorageBackend::get_root_metadata` / `get_all_root_metadata` — read views that incorporate pending writes and hide tags of zero-count roots.
- A root holds one tag (last write wins); the tag is deleted automatically when the root count drops to zero.

**Durability.** Tags are staged in memory and committed via `Update::SetRootMetadata` / `DeleteRootMetadata` in the same `DB::batch_update` as the key's persist count — `batch_update` is atomic over its whole iterator (SQLite: one transaction; ParityDb: one `commit_changes`), and every batch that writes a tag also carries that key's count, so a crash can never make one durable without the other. A pending tag flushes at the first flush (full or eviction-driven) at which its key's count is durable. A crash mid-pruning leaves the on-disk tags intact, so re-submitting the same values retries the release; a crash after the flush makes re-submission a no-op. Unpersisting itself does no metadata I/O — all metadata decisions are deferred to flush, where they batch.

**Storage.**

- SQLite: new `root_metadata` table, created with `IF NOT EXISTS` — existing files upgrade transparently.
- ParityDb: uses column 2, reserved for GC purposes since layout v2. `NUM_COLUMNS` is unchanged, so existing databases (including embedded, column-offset instances) stay compatible.
- Implemented for `InMemoryDB`, `SqlDB`, `ParityDb`, and the `WrappedDB` pass-through.

**Testing.** Per-backend metadata CRUD across all three backends; the end-to-end pruning workflow (tag at persist → release a batch of tags in one call → repeat is a no-op → GC collects exactly the released roots); multi-persist lifecycle (tag survives until the count reaches zero); flush atomicity (staged view visible immediately, disk unchanged until flush, crash-retry handle intact in between); deferred tagging after a durable persist; eviction-flush timing (a tag never flushes while its count delta is still cached, but flushes with unrelated evictions once its count is durable).

## Sanity Checklist

This PR:

### DIFF
--- a/CHANGELOG_storage_core.md
+++ b/CHANGELOG_storage_core.md
@@ -3,6 +3,7 @@
 ## Version `1.2.0`
 
 - feat: add incremental garbage collector, running in a time-bounded way. This requires databases to support a new scan operation.
+- feat: add GC-root metadata (behind `gc-v1`): `Sp::persist_with_metadata` / `StorageBackend::persist_with_metadata` attach an opaque value (e.g. a block hash) when the tag is known at persist time; `StorageBackend::set_root_metadata` tags an already-persisted root without changing its persist count. `StorageBackend::unpersist_by_metadata` later releases matching roots given only those values -- zero-clamped, in one backend borrow, with no caller-side index. Metadata is staged in the write cache and committed in the same `batch_update` as any persist-count change, and deleted on flush when the count drops to zero. Backed by a new `root_metadata` table in SQLite, and the previously reserved column 2 in ParityDb.
 - feat: allow parityDB to use existing instance
 - fix: removed race condition from `force_as_arc`
 - fix: prevent a panic in `Sp` serialization with a mix of 'promoted' and 'unpromoted' keys.
--- a/storage-core/src/arena.rs
+++ b/storage-core/src/arena.rs
@@ -475,6 +475,10 @@ type MetaData<D> = HashMap<ArenaHash<<D as DB>::Hasher>, Node>;
 /// need to include the type in the key to avoid collisions in some cases.
 type DynTypedArenaHash<H> = (ArenaHash<H>, TypeId);
 
+/// Map from content hash to the child-key representation used while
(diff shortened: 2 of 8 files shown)


tkerber:
~Not mergable as-is; this is a backwards incompatible layout change, and therefore would need a new layout version.~

My mistake, I need to learn to read. That said, I'm not a fan of this, as I explained in slack. I don't want to layer a key-value store on ledger storage.

Issue #723, opened 2026-08-21
CI: Nix store cache is purged every run, so aarch64/macOS legs rebuild the toolchain and crash


chrisferry:
The `build-ledger` and `build-proof-server` jobs fail intermittently on the `ubuntu-24.04-arm` and `macos-latest` legs with:

```
building '/nix/store/...-rust-stable-with-std-2026-04-16.drv'...
building '/nix/store/...-cargo-nightly-complete-2025-11-20.drv'...
error: Nix daemon disconnected unexpectedly (maybe it crashed?)
```

Which job fails varies per attempt, and re-running does not reliably clear it.

## Why it never self-heals

`.github/workflows/ci.yml` configures the store cache as:

```yaml
- name: Restore and save Nix store
  uses: nix-community/cache-nix-action@... # v7
  with:
    primary-key: nix-${{ runner.os }}-${{ hashFiles('**/*.nix', '**/flake.lock') }}
    restore-prefixes-first-match: nix-${{ runner.os }}-
    purge: true
    purge-prefixes: nix-${{ runner.os }}-
    purge-created: 0
```

`purge-created: 0` purges every cache matching the prefix that was created more than **0 seconds** ago — i.e. all of them, on every run. The action only *saves* in its post step on success. So a failing job purges the caches and saves nothing, and the next run starts cold again. Observed directly on the macOS leg:

```
primary-key: nix-macOS-5dc352f8032674a98d4515fba71b7bb5162c6ab2302ddcc89b6209c1c10c0b3a
Could not find a cache with the given "primary-key" and "paths".
```

With a cold store, that job built **146 derivations from source** (805 paths came from substituters). The heavy ones are the toolchain derivations, and that is where the daemon dies.

This also explains the platform split: only the multi-platform matrices fail. Every `ubuntu-latest`-only job passes, and `extra-substituters = https://cache.iog.io` is x86_64-linux-centric, so the aarch64-linux and aarch64-darwin legs have no prebuilt substitutes for these derivations and must build them locally.

## The real error is being hidden

```yaml
run: |
  cargo install cargo2junit
  nix develop .#ci -c cargo test --release -- ... | tee test-results.json
  jq -cR 'fromjson? | select( .type != "report" )' test-results.json | cargo2junit > ./target/test-results.xml
```

[...]

PR #724, opened 2026-08-21
fix(ci): stop purging the Nix store cache on every run

chrisferry:
Implements the first two fixes from #723. The `build-ledger` / `build-proof-server` legs on `ubuntu-24.04-arm` and `macos-latest` crash with `error: Nix daemon disconnected unexpectedly`, and re-running never clears it. Here is the mechanism, verified against the pinned action's `action.yml`:

1. The restore step runs and purges caches. `purge-created: 0` means *purge every selected cache created more than zero seconds ago* — i.e. all of them, every run.
2. The job fails. `cache-nix-action` declares `post-if: "success()"`, so the save step never runs and nothing is written back.
3. The next run finds nothing. Observed directly on the macOS leg:
   ```
   primary-key: nix-macOS-5dc352f8032674a98d4515fba71b7bb5162c6ab2302ddcc89b6209c1c10c0b3a
   Could not find a cache with the given "primary-key" and "paths".
   ```
4. Cold store, so that job built **146 derivations from source** (805 came from substituters). The expensive ones are the toolchain derivations, and that is where the daemon dies.

It is a closed loop: purge always, save only on success. That also explains the platform split — every `ubuntu-latest`-only job passes, while `extra-substituters = https://cache.iog.io` is x86_64-linux-centric so the aarch64 legs have no substitutes for these derivations and must build them.

**Change 1 — keep the cache we depend on.** `purge-primary-key: never` and `purge-created: P7D`, applied to all five cache blocks. Purging still garbage-collects, but never the current key and never anything from the past week.

**Change 2 — stop hiding the real error.** `set -eo pipefail` on the two test steps. `nix develop .#ci -c cargo test … | tee test-results.json` returned tee's exit status, so the step continued past a failed build and died later at `> ./target/test-results.xml`. Every infrastructure crash was therefore reported as `./target/test-results.xml: No such file or directory`, which is why this took a while to localise.

Verified: `actionlint` findings unchanged at 42 before and after (all pre-existing `SC2086`/`SC2129` in untouched steps), `yamllint` clean.

**Residual risk worth naming:** because saving is still gated on `success()`, the aarch64/macOS caches only warm up once one of those legs goes green. If they still cannot finish a cold build, the remaining items in #723 apply — capping Nix parallelism, reclaiming runner disk, larger runners, or publishing the toolchain derivations to an org Cachix for those platforms. I deliberately did not bundle that speculative tuning here.

Refs #723

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -61,7 +61,7 @@ jobs:
             accept-flake-config = true
 
       - name: Restore and save Nix store
-        uses: nix-community/cache-nix-action@v7
+        uses: nix-community/cache-nix-action@7df957e333c1e5da7721f60227dbba6d06080569 # v7
         with:
           primary-key: nix-${{ runner.os }}-${{ hashFiles('**/*.nix', '**/flake.lock') }}
           restore-prefixes-first-match: nix-${{ runner.os }}-
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -67,7 +67,8 @@ jobs:
           restore-prefixes-first-match: nix-${{ runner.os }}-
           purge: true
           purge-prefixes: nix-${{ runner.os }}-
-          purge-created: 0
+          purge-primary-key: never
+          purge-created: P7D
 
       - name: Configure cargo
         run: .github/scripts/configure-cargo.sh
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,8 @@ jobs:
           restore-prefixes-first-match: nix-${{ runner.os }}-
           purge: true
           purge-prefixes: nix-${{ runner.os }}-
-          purge-created: 0
+          purge-primary-key: never
+          purge-created: P7D
 
       - name: Configure cargo
         run: .github/scripts/configure-cargo.sh
@@ -124,7 +125,8 @@ jobs:
           restore-prefixes-first-match: nix-${{ runner.os }}-
           purge: true
           purge-prefixes: nix-${{ runner.os }}-
-          purge-created: 0
+          purge-primary-key: never
+          purge-created: P7D
 
       - name: Configure cargo
         run: .github/scripts/configure-cargo.sh
@@ -194,7 +196,8 @@ jobs:
           restore-prefixes-first-match: nix-${{ runner.os }}-
           purge: true
           purge-prefixes: nix-${{ runner.os }}-
-          purge-created: 0
+          purge-primary-key: never
+          purge-created: P7D
 
       - name: Configure cargo
         run: .github/scripts/configure-cargo.sh
@@ -217,6 +220,7 @@ jobs:
 
       - name: Build and Test (default features)
         run: |
+          set -eo pipefail
           cargo install cargo2junit
           nix develop .#ci -c cargo test --release -- -Z unstable-options --format json --report-time --nocapture --test-threads=1 | tee test-results.json
           # Filter is a workaround for johnterickson/cargo2junit#85
@@ -256,7 +260,8 @@ jobs:
           restore-prefixes-first-match: nix-${{ runner.os }}-
           purge: true
           purge-prefixes: nix-${{ runner.os }}-
-          purge-created: 0
+          purge-primary-key: never
+          purge-created: P7D
 
       - name: Configure cargo
         run: .github/scripts/configure-cargo.sh
@@ -279,6 +284,7 @@ jobs:
 
       - name: Build and Test - Proof server (default features)
         run: |
+          set -eo pipefail
           cargo install cargo2junit
           cd proof-server
           nix develop .#ci -c cargo test --release -- -Z unstable-options --format json --report-time --nocapture --test-threads=1 | tee test-results-proof-server.json
--- a/.github/workflows/docker-push.yml
+++ b/.github/workflows/docker-push.yml
@@ -134,7 +134,7 @@ jobs:
           docker push ghcr.io/midnight-ntwrk/proof-server:$PROOF_SERVER_VERSION-arm64
 
   docker-manifest:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     needs: [docker-amd64, docker-arm64]
     permissions:
       contents: read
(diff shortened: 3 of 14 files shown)


chrisferry:
Pushed a second commit — **my first pass was incomplete and would not have worked.**

Fixing `ci.yml` alone is insufficient because every workflow in this repo purges with the same `nix-${{ runner.os }}-` prefix. So `nightly.yml`, `docker-push.yml` (×3), `benchmark.yml`, `prerelease.yml` (×3), `publish.yml`, `test-coverage.yml`, `prepare-release.yml`, `prepare-cost-model-docker.yml` and `upload-srs.yml` would each still have deleted the cache `ci.yml` had just saved. Thirteen further occurrences, now all fixed — `purge-created: 0` no longer appears anywhere in `.github/workflows/`.

`nightly.yml` is the clearest illustration of why this matters. It is the one build that reliably passes on `ledger-8`, because it runs on `ubuntu-latest-8-core-x64` rather than a standard runner — so it is the natural candidate to seed a warm Linux cache. But it shares the primary key `nix-${{ runner.os }}-${{ hashFiles(...) }}` with `ci.yml` **and** purged everything on every run, so its successes were destroying exactly the cache PR builds needed.

Verified per file: `actionlint` finding counts unchanged (7→7, 18→18, 6→6, 2→2, 0→0, 0→0, 7→7, 6→6), and the only `yamllint` output is a pre-existing `too many spaces before colon` at untouched lines in `nightly.yml:17` and `benchmark.yml:15`, identical before and after.

One thing I deliberately left out: six of these files reference `nix-community/cache-nix-action@v7` as a **mutable tag** (`nightly.yml`, `docker-push.yml` ×3, `benchmark.yml`, `prepare-cost-model-docker.yml`, `upload-srs.yml`) while the other files pin the same version to `7df957e…`. Worth pinning for supply-chain hygiene, but it is unrelated to this failure and I did not want to widen the diff.

chrisferry:
Third commit: pinned the remaining mutable action references, since this is going in via force-merge and the hygiene fix belongs with it rather than in a follow-up nobody opens.

Eight references were still on mutable tags while the rest of the repo pinned SHAs:

| File | Ref | Now |
|---|---|---|
| `nightly.yml`, `benchmark.yml`, `prepare-cost-model-docker.yml`, `upload-srs.yml`, `docker-push.yml` (×3) | `cache-nix-action@v7` | `@7df957e333c1e5da7721f60227dbba6d06080569 # v7` |
| `upload-srs.yml` | `configure-aws-credentials@v6` | `@e6de054238d6b7531b4efff3b6587d9aade6a06c # v6.2.3` |

Two things I checked rather than assumed:

- **`cache-nix-action@v7` currently resolves to `7df957e…`** — the exact SHA the other files already pin. So this is behaviourally a **no-op**: no version moves, and all eighteen call sites now reference one identical SHA instead of a mix of pinned and floating.
- **`configure-aws-credentials@v6` is an annotated tag object**, so the ref API returns `451ce2a…`, which is the *tag* object and not usable in `uses:`. Dereferenced it to the underlying commit `e6de054…` (release 6.2.3). Pinning the tag SHA there would have broken the workflow.

`actionlint` counts unchanged on every touched file (7→7, 18→18, 6→6, 6→6, 6→6), and the only `yamllint` output remains the pre-existing `too many spaces before colon` at untouched lines.

The branch is now three commits: the `ci.yml` cache fix, the same fix across the other nine workflows, and these pins. `purge-created: 0` and mutable action tags are both gone from `.github/workflows/` entirely.

chrisferry:
Fourth commit: pinned the runner images — but **only the ones that are safe to pin without an admin change**, and the reason matters.

Verified from a real run rather than assumed, so these are no-ops today:

```
ubuntu-latest  ->  Image: ubuntu-24.04
macos-latest   ->  Image: macos-26-arm64
```

So pinning changes nothing now; it just stops GitHub migrating the image under us later. `ubuntu-24.04` was already pinned in three places here, so it is also the consistent target. **18 occurrences pinned across 8 workflows.**

### What I deliberately did not pin, and why

**`ci.yml` is untouched.** Its labels live in matrix `os:` arrays, so the job names carry the runner — and `ledger-8` branch protection requires these contexts (`strict: true`):

```
check (ubuntu-latest)               build-ledger (ubuntu-latest)
build-wasm (ubuntu-latest)          build-ledger (macos-latest)
build-proof-server (ubuntu-latest)  integration-tests (ubuntu-latest)
build-ledger (ubuntu-24.04-arm)     build-proof-server (ubuntu-24.04-arm)
scan                                license/cla
```

Seven of those embed `ubuntu-latest` / `macos-latest`. Pinning the matrix renames them, the required checks would never report again, and **every PR in the repo would block indefinitely** — a worse outage than the one this PR fixes. It needs the workflow change and the protection contexts updated together, which is an admin operation and belongs in its own change.

`ci.yml` itself says as much, which is good corroboration:

```yaml
# Inline matrix job (not a reusable-workflow call) so the check name stays
# `integration-tests (ubuntu-latest)`.
```

**`ubuntu-latest-8-core-x64` and `ubuntu-latest-4-core-arm64` are untouched.** Despite the name, those are custom larger-runner labels, not GitHub images — `actionlint` flags them as unknown labels. Renaming them would leave those jobs unschedulable.

`actionlint` counts unchanged on every touched file (18→18, 2→2, 7→7, 0→0, 2→2, 0→0, 0→0, 6→6).

### Follow-up for the matrix pin

When someone wants it, the workflow edit has to land together with:

```sh
gh api -X PATCH repos/midnightntwrk/midnight-ledger/branches/ledger-8/protection/required_status_checks \
  -f "contexts[]=check (ubuntu-24.04)" \
  -f "contexts[]=build-wasm (ubuntu-24.04)" \
  -f "contexts[]=build-ledger (ubuntu-24.04)" \
  -f "contexts[]=build-ledger (macos-26)" \
  -f "contexts[]=build-ledger (ubuntu-24.04-arm)" \
  -f "contexts[]=build-proof-server (ubuntu-24.04)" \
  -f "contexts[]=build-proof-server (ubuntu-24.04-arm)" \
  -f "contexts[]=integration-tests (ubuntu-24.04)" \
  -f "contexts[]=scan" -f "contexts[]=license/cla"
```

Worth noting for #723: `macos-latest` now resolves to **macos-26-arm64**. If that migration is recent it is a plausible contributor to the Nix daemon crashes, and pinning the matrix to a known-good older image (e.g. `macos-15`) would be a cheap way to test that theory.

PR #732, opened 2026-08-30
ci: save Nix store on failure and cap parallelism for aarch64/macOS

### LINKED ISSUE #723: CI: Nix store cache is purged every run, so aarch64/macOS legs rebuild the toolchain and crash

Ansh-Sonkusare:
Closes #723.

PR #724 fixed the primary cause (`purge-created: 0` → `P7D`) and `pipefail`, but its author noted a residual risk:

> because saving is still gated on `success()`, the aarch64/macOS caches only warm up once one of those legs goes green. If they still cannot finish a cold build, the remaining items in #723 apply.

This PR addresses that residual risk in `build-ledger` and `build-proof-server` (the failing multi-platform jobs).

## Changes

**Save on failure** — split the combined `cache-nix-action` step into separate restore + save steps. Restore runs unconditionally; save runs with `if: always()` using the `cache-nix-action/save` sub-action. A failing cold build now writes whatever store it managed to populate, so the next attempt can restore a partial cache instead of starting from zero. This breaks the death spiral: cold build fails → nothing saved → next run cold → fails again.

**Cap Nix parallelism** — add `max-jobs = 2` and `cores = 2` to `extra_nix_config` for both jobs. Reduces peak memory pressure during cold toolchain builds on the smaller aarch64/macOS runners, lowering the risk of the Nix daemon being killed before the first warm cache can be established.

## Test plan

- [ ] A failing `build-ledger` or `build-proof-server` run still writes a Nix store cache
- [ ] The subsequent run reports a cache hit rather than `Could not find a cache`
- [ ] `build-ledger` and `build-proof-server` pass on all matrix legs without re-runs

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tkerber (changes requested):
Thank you for your contribution! We definitely have a bit of a struggle with caching right now, so this is very welcome. Full disclosure however: I didn't feel like thinking about cache behaviour right now, so I kicked this one to Claude. This highlighted some pre-existing issues (not on you, but also would be nice to fix), and some newly created ones:


>  1. ci.yml:229, :303 — the save permanently pins a partial cache. With an immutable primary key and the default purge-primary-key: never, saveImpl.ts returns early on an exact key hit and purgeCachesByPrimaryKeyAndPrefixes filters the primary cache out of the purge set. So a failed cold run writes a half-populated /nix under nix-<os>-<hash>, and every later run — green ones included — restores that partial store and skips saving. Frozen until flake.lock or a *.nix changes. Previously post-if: success() meant only a complete store could claim the key. Fix: append -${{ github.run_id }} to the save key and rely on restore-prefixes-first-match, or set purge-primary-key: always.
>  2. ci.yml:197, :270, :48, :124 — the key uses runner.os, not runner.arch. runner.os is Linux for both ubuntu-latest and ubuntu-24.04-arm, and hashFiles is arch-independent. Five Linux legs share the key nix-Linux-<hash>; whichever finishes first claims it, and the aarch64 legs restore a wrong-arch multi-GB store they can never replace. This is the exact "aarch64 caches never warm up" symptom the PR targets, and the change leaves it intact. Add -${{ runner.arch }} to primary-key and restore-prefixes-first-match.
>  3. ci.yml:231, :305 — always() fires on cancellation. concurrency.cancel-in-progress: true (line 12) cancels the running job on every push, so a run cancelled minutes into a cold build uploads a near-empty store and, per finding 1, claims the key permanently. Cache poisoning becomes the common case. Use if: '!cancelled()'.
>  4. ci.yml:191, :264 — max-jobs/cores miss the memory hog. They bound derivation builds, not cargo/rustc inside nix develop .#ci -c cargo test --release, which picks parallelism from the visible CPU count. The cap also applies to the healthy x86_64 leg, lengthening its cold builds for no benefit. Scope it to the ARM/macOS legs and constrain cargo with CARGO_BUILD_JOBS.
>  5. ci.yml:303 — save step precedes Upload Artifact in build-proof-server. A timeout or cancellation during the multi-GB upload loses the test-results-* artifacts. build-ledger orders it correctly, last; always() makes ordering free here.

On balance, the cancellation handling & cache overwrite, together with the fact that we _do_ end up with frequent cancelled jobs, seems like a blocker to me – but I would like to see this fixed up!

PR #734, opened 2026-09-02
perf(serialize): decode integers without a 67-octet buffer

gilescope:
`via_scale!` routed every `u32`/`u64`/`u128` through `ScaleBigInt`, whose
`serialized_size` scans 67 octets backwards to find a width the type already
knows — about 260 octets moved to decode eight. `scale_size`/`scale_write`/
`scale_read` compute it instead. `ScaleBigInt` is untouched.

Profiled by instruction count rather than wall-clock, `serialized_size` alone was
19.45% of every node decoded. On a cpu it is a few hundred octets of `memcpy`,
which is why no one noticed but all it adds up.

### DIFF
--- a/serialize/src/util.rs
+++ b/serialize/src/util.rs
@@ -712,3 +712,146 @@ use crate as serialize;
 
 #[cfg(feature = "proptest")]
 randomised_serialization_test!(String);
+
+#[cfg(test)]
+mod scale_compat {
+    //! **The gate on any change to how integers serialise.**
+    //!
+    //! `ScaleBigInt` is the reference: the general path that every `via_scale!` type went
+    //! through when this was written. Anything faster has to agree with it *byte for byte*,
+    //! and this compares them directly rather than against a golden file, so the oracle
+    //! cannot drift out of date.
+    //!
+    //! ⚠︎ Byte-identical is not a nicety here. Storage is content-addressed — a node's key
+    //! *is* the hash of its serialisation — so a changed encoding changes every node hash,
+    //! every parent that references it, and the state root. That is a migration of the whole
+    //! ᴅᴀɢ, not a version bump.
+    use super::*;
+
+    /// Every value where the encoding changes width, either side of each boundary, plus the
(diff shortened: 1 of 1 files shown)


evilrobot-01 (commented):
Identified the same issue through independent indexer profiling. Given the context of my Opus performance session, I switched to Fable and asked for a review.

> - It removes exactly our two hottest leaves: `serialized_size` at 17% (1,408 samples) and the `ScaleBigInt` decode at 12% (1,002 samples) of on-CPU. Their 19.45% instruction-count figure independently corroborates our sample-based number.
> - It also takes pressure off two adjacent findings: the 67-byte construct-and-scan was part of our `memset`/`copy_nonoverlapping` cost, and `DirectChildNode::new` recomputing sizes during deserialization (storage-core, untouched by this PR) drops from "dominant leaf" to near-free once sizing is arithmetic.

Correctness was verified, but it was noted that the error message changed which might be worth considering for a like-for-like replacement. An additional recommendation was to increase of test coverage as follows:

> One property worth being conscious of: the decoder — original and new alike — accepts values in `[2^24, 2^30)` in *both* the 4-byte form and the n-byte form (4 value bytes, top byte nonzero passes the only check). Reproducing that dual acceptance is the right call for consensus compatibility, but it's currently pinned by nothing.

> The `scale_compat` tests prove **encoder** equivalence thoroughly (the corpus of every width boundary ±1 is exactly right) but decode is only tested via round-trips of canonical encodings plus four handwritten non-canonical streams. A differential decode test — arbitrary byte strings fed to both `ScaleBigInt::deserialize → TryFrom` and the new path, asserting identical accept/reject, value, and bytes consumed — would pin the behaviour the handwritten cases can't reach, including the `[2^24, 2^30)` window above. The crate already has proptest infrastructure (`randomised_serialization_test!`), so it's cheap to add.

gilescope:
The other thing that concerns me is that we allow non-canonical representations of numbers. It would be probably better for security if there were only one way to represent a number.

PR #738, opened 2026-09-02
Iquerejeta/verify proof clean

iquerejeta:
## Description

Implements verify proof instruction in ZKIR. 

Temporarily opening the full PR here. Will eventually split it between this and the https://github.com/midnightntwrk/midnight-zkir repo. 

This work replaces https://github.com/midnightntwrk/midnight-ledger/pull/661. 

The last commit, the one bumping versions, was implemented with heavy assistance of AI. 

## What this PR adds

Two new ZKIR instructions in `zkir-v3`:

- **`inner_proof { guard, output }`** — binds `output` to the next inner proof from the prover's witness (`ProofPreimage::inner_proofs`), consumed positionally: one per instruction whatever its guard, so the witness vector's length is fixed by the circuit and not by the path taken. In-circuit it is a free prover witness carrying no constraints of its own.
- **`verify_proof { guard, vk_hash, instance, proof }`** — verifies that proof against the verifying key named by `vk_hash`, over `instance` as its public inputs. No outputs.

The final KZG pairing is **not** performed in-circuit. The circuit does the cheap part of the verification and exposes the resulting **deferred accumulator** as public inputs; the outer verifier reconstructs it and runs the pairing that discharges it. That split is the core of the design.

### `zkir-v3/src/ir_instructions/verify_proof.rs`

The two halves of the instruction, built on the verifier-gadget primitives midnight-circuits exposes:

- `verify_proof_offcircuit` runs `plonk::prepare` and returns the `Accumulator` the check defers. There is no verdict to return.
- `verify_proof_incircuit` emits the same thing as constraints via `ZkStdLib`'s verifier, constraining exactly the fields the other returns.

Guard semantics differ by necessity. Guarded off, the off-circuit half returns the trivial accumulator without reading key or proof — required, not an optimisation, since a guarded-off `inner_proof` binds the empty blob on which `plonk::prepare` fails at EOF. The in-circuit half always prepares, because a circuit's shape cannot depend on a witness, and reduces the accumulator to the trivial one so the deferred pairing holds whatever the prover supplied.

### `zkir-v3/src/decider.rs`

`DeciderKind` declares, per inner verifying key, what obligation an inner proof carries beyond a plain PLONK verification — `None`, or `Collapsed` for one fully-collapsed, fixed-base-resolved accumulator in the last `accumulator_pi_len` fields of its instance. IVC chain proofs are the motivating case.

The tag is a single consensus byte written by declaration order, so variants may only be appended. **Declaring `DeciderKind::None` for an inner proof that does have a deferred accumulator is a silent soundness bug** — it is the verifier's responsibility to declare it correctly.

`serialize_vk` / `deserialize_vk` encode `[tag] | MidnightVK::write(Processed)`. Trailing bytes are refused: the blob's digest names the key's fixed bases, so a padded blob would give the same key a second identity and the circuit a second shape.

### `transient-crypto/src/proofs.rs`

- `Proof` moves from the tuple struct `Proof(Vec<u8>)` to `Proof { bytes, accumulators }`, one accumulator per `verify_proof` instruction in the outer circuit, in instruction order.
- `DeferredAccumulator` stores only two points. An exposed accumulator has been through `resolve_fixed_bases` and `collapse`, leaving each side a single base whose scalar is constrained to one, so the always-one scalars are regenerated rather than stored — 384 bytes down to 96. A point off the curve or outside the prime-order subgroup is refused at deserialization, rather than after the proof verifies.
- `VerifierKey::verify` and the batch path check each carried accumulator's pairing after the outer proof verifies, each on its own rather than subsumed by the first.
- `VerifierKey::mock_verify` now takes the `Proof` alongside the statement, so calibrated-mock timings account for the deferred pairings too.
- `ProofPreimage` gains `inner_proofs: Vec<InnerProofWitness>`.

### `zkir-v3/src/ir_vm.rs`

`validate_inner_proofs` rejects a malformed pairing before proving or key generation: each `verify_proof` must name a proof some earlier `inner_proof` bound, under the same guard operand, and each bound proof must be used exactly once. Guards must match because each instruction reads only its own — guarded off, `inner_proof` binds an empty blob that a `verify_proof` guarded on would then fail to verify.

## Versioning and compatibility

Most of the file count comes from here. The formats `verify_proof` changed are versioned so that everything already written still decodes.

| Type | Was | Now |
| --- | --- | --- |
| `Proof` | `proof[v5]` | `proof[v6]` |
| `ProofPreimage` | `proof-preimage` | `proof-preimage[v2]` |
| `ProofPreimageVersioned` | `proof-preimage-versioned` | `proof-preimage-versioned[v2]` |
| ZKIR `Instruction` | `ir-instruction[v3]` | `ir-instruction[v4]` |
| `VerifierKey` | `verifier-key[v7]` | `verifier-key[v8]` |
| new | — | `deferred-accumulator`, `inner-proof-witness` |

- `ProofVersioned` gains `V4(Proof)` at discriminant 3, carrying `proof[v6]`. `V2` and `V3` are retyped to `transient_crypto_old::proofs::Proof`, which is the `proof[v5]` layout they always had on the wire, so every already-written transaction still deserializes and the `proof-versioned` tag itself is unchanged.
- `IrSource` gains `verify_proof_vks: Vec<Vec<u8>>` behind a new `IrMinorVersion::V1`, now the default. Its `Serializable` is hand-written and branches on the version: `V0` blobs keep their exact byte layout and still load, and serializing a `V0` that carries key material is an error rather than a silent drop.
- A proof's version is now taken from the tag of the verifier key it was proven against (`ProofVersioned::of_verifier_key_tag`), in both `ledger::prove` and the proof server, instead of being hardcoded — so a tag bump in either proof system carries over rather than going stale.
- A verifier key in `ContractOperation`'s `v3` slot is now `verifier-key[v8]` against `proof[v6]`. The key's own bytes are unchanged, so a stored contract operation still decodes.
- `zkir`'s `preimage_to_v1` refuses a preimage carrying inner proofs: the v1 pipeline has no `verify_proof` to consume them.
- Version bumps: `transient-crypto` 4.0.0, `onchain-runtime` 5.0.0-alpha.1.

## Sanity Checklist

This PR:
- [x] contains changes to architecture [^1]
- [x] contains breaking API changes [^1][^2]
- [x] contains data format changes [^1]


zkir-v3/src/decider.rs
iquerejeta, line 1: Note - this file might belong in transient-crypto.

ledger/src/construct.rs
```diff
@@ -559,6 +559,7 @@ impl<D: DB> ContractCallExt<D> for ProofPreimage {
         let binding_input = 0u8.into();
 
         let proof = ProofPreimage {
+            proof_witnesses: vec![],
```
miguel-ambrona: Why is `proof_witnesses` in the first position? I would respect the `ProofPreimage` struct order.

ledger-wasm/src/tx.rs
```diff
@@ -503,10 +503,10 @@ impl Transaction {
                 Ok(tagged_deserialize(&mut &result[..]).or_else(|_| {
                     let ppiv = tagged_deserialize::<ProofVersioned>(&mut &result[..])?;
                     match ppiv {
-                        ProofVersioned::V2(proof) => Ok::<_, std::io::Error>(proof),
+                        ProofVersioned::V4(proof) => Ok::<_, std::io::Error>(proof),
```
tkerber: Backwards compat: We want to support old proofs? Also, what happened to v3?

ledger/src/construct.rs
tkerber: This path being stubbed will still block meaningful usage.

ledger/src/dust.rs
```diff
@@ -2173,7 +2174,7 @@ mod tests {
         };
         let binding = Pedersen::from(rng.r#gen::<PedersenRandomness>());
         let proven_dust_spend = dust_spend.prove(prover, 0, binding).await.unwrap();
-        assert_eq!(proven_dust_spend.proof.0.len(), DUST_SPEND_PROOF_SIZE);
+        assert_eq!(proven_dust_spend.proof.bytes.len(), DUST_SPEND_PROOF_SIZE);
```
tkerber: `proof.serialized_size()` is a more reasonable proxy imo

ledger/src/prove.rs
```diff
@@ -373,8 +373,10 @@ impl<D: DB> ContractCall<ProofPreimageMarker, D> {
                 let tag = peek_tag(&mut std::io::Cursor::new(&vk))
                     .map_err(TransactionProvingError::Tokio)?;
                 match tag.as_str() {
-                    "verifier-key[v6]" => ProofVersioned::V2(proof),
-                    "verifier-key[v7]" => ProofVersioned::V3(proof),
+                    "verifier-key[v6]" => {
+                        ProofVersioned::V2(transient_crypto_old::proofs::Proof(proof.bytes))
+                    }
+                    "verifier-key[v7]" => ProofVersioned::V4(proof),
```
tkerber: Should probably be v8?

ledger/src/structure.rs
```diff
@@ -227,7 +227,7 @@ impl<D: DB> PedersenDowngradeable<D> for Pedersen {
 
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Storable)]
 #[storable(base)]
-#[tag = "proof-preimage-versioned"]
+#[tag = "proof-preimage-versioned[v2]"]
```
tkerber, line 234: This should probably *not* move *but* have `ProofPreimageVersioned` have an old (V2) variant and a new (V3) variant. We already had half of the machinery for the proof preimage change!

ledger/Cargo.toml
```diff
@@ -23,7 +23,7 @@ coin-structure = { version = "3.0.0", path = "../coin-structure", package = "mid
 onchain-runtime = { version = "4.0.0", path = "../onchain-runtime", package = "midnight-onchain-runtime" }
 zswap = { version = "9.0.0", path = "../zswap", package = "midnight-zswap" }
 base-crypto = { version = "1.1.0", path = "../base-crypto", package = "midnight-base-crypto" }
-transient-crypto = { version = "3.0.0", path = "../transient-crypto", package = "midnight-transient-crypto" }
+transient-crypto = { version = "4.0.0", path = "../transient-crypto", package = "midnight-transient-crypto" }
```
tkerber, line 26: Backwards compat: We really need all three loaded. The naming was bad from the start, should have been `transient_crypto_2`, `transient_crypto_3`, and `transient_crypto_4`. We don't have the luxury of dropping old validation behaviour.

onchain-runtime-wasm/src/primitives.rs
```diff
@@ -254,6 +254,7 @@ pub fn proof_data_into_serialized_preimage(
     input.value_only_field_repr(&mut comm_comm_preimage);
     output.value_only_field_repr(&mut comm_comm_preimage);
     let preimage = transient_crypto::proofs::ProofPreimage {
+        proof_witnesses: vec![],
```
tkerber: Again, this needs un-stubbing for practical usage (matching ledger/src/construct.rs)

onchain-state/Cargo.toml
```diff
@@ -12,7 +12,7 @@ public-internal-structure = []
 
 [dependencies]
 coin-structure = { version = "3.0.0", path = "../coin-structure", package = "midnight-coin-structure" }
-transient-crypto = { version = "3.0.0", path = "../transient-crypto", package = "midnight-transient-crypto" }
+transient-crypto = { version = "4.0.0", path = "../transient-crypto", package = "midnight-transient-crypto" }
```
tkerber, line 15: As with `ledger/Cargo.toml`, we need all three versions of verifier keys :/

proof-server/src/endpoints.rs
```diff
@@ -306,7 +306,7 @@ pub(crate) async fn prove(
                             .map_err(WorkError::BadInput)?
                             .0;
 
-                        ProofVersioned::V2(proof)
+                        ProofVersioned::V4(proof)
```
tkerber: Proof server still needs to handle old version keys :/

Or am I misreading here, and the new proofs work with the old keys?

transient-crypto/src/proofs.rs
```diff
@@ -129,13 +134,35 @@ lazy_static! {
...
+/// `bytes` is the raw PLONK proof produced by midnight-zk. `accumulators`
+/// carries one deferred KZG accumulator per `verify_proof` instruction in the
+/// outer circuit, encoded as its public-input field elements
+/// (`accumulator_pi_len()` `Fr`s per block).
 #[cfg_attr(feature = "proptest", derive(Arbitrary))]
 #[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Serializable, Storable)]
 #[storable(base)]
-#[tag = "proof[v5]"]
-pub struct Proof(pub Vec<u8>);
+#[tag = "proof[v6]"]
```
tkerber, line 146: Nice!

zkir-v3/src/decider.rs
tkerber: I'm not sure it does – it's not on the verifier path, and it's optional for a prover, or am I wrong?

zkir-v3/Cargo.toml
```diff
@@ -34,7 +34,7 @@ proptest = [
 [dependencies]
 base-crypto = { version = "1.1.0", path = "../base-crypto", package = "midnight-base-crypto" }
 serialize = { version = "1.0.0", path = "../serialize", package = "midnight-serialize" }
-transient-crypto = { version = "3.0.0", path = "../transient-crypto", package = "midnight-transient-crypto", features = [
+transient-crypto = { version = "4.0.0", path = "../transient-crypto", package = "midnight-transient-crypto", features = [
```
tkerber, line 38: backwards compat: zkir 3.0 is stable, so we need to support keys made with it -> probably need to pull in both versions.

zkir/Cargo.toml
```diff
@@ -26,7 +26,7 @@ proptest = [
 [dependencies]
 base-crypto = { version = "1.1.0", path = "../base-crypto", package = "midnight-base-crypto" }
 serialize = { version = "1.0.0", path = "../serialize", package = "midnight-serialize" }
-transient-crypto = { version = "3.0.0", path = "../transient-crypto", package = "midnight-transient-crypto", features = ["cli"] }
+transient-crypto = { version = "4.0.0", path = "../transient-crypto", package = "midnight-transient-crypto", features = ["cli"] }
```
tkerber, line 29: As before – zkir v2 is stable (& mature / nearing deprecation), and probably shouldn't be jumping versions.

zswap/src/prove.rs
```diff
@@ -245,12 +245,12 @@ mod tests {
         let inp =
             Input::new_contract_owned(&mut rng, &qcoin, None, Default::default(), &tree).unwrap();
         let inp_proven = inp.prove(provider.split()).await.unwrap();
-        assert_eq!(inp_proven.proof.0.len(), INPUT_PROOF_SIZE);
+        assert_eq!(inp_proven.proof.bytes.len(), INPUT_PROOF_SIZE);
```
tkerber: As in Dust, this should probably use the serialized size for accuracy.

tkerber (changes requested):

transient-crypto/src/proofs.rs
```diff
@@ -696,6 +696,29 @@ pub trait ProvingProvider {
...
+    PartialEq,
+    Eq,
+    PartialOrd,
+    Ord,
+    Serializable,
+    Hash,
+    Storable,
+    Zeroize,
+    ZeroizeOnDrop,
+)]
```
miguel-ambrona, line 972: Do we really need all of these?

ledger/src/construct.rs
miguel-ambrona: This happens in more than one place.

transient-crypto/src/proofs.rs
```diff
@@ -359,6 +364,39 @@ impl<T: Zkir> Deserializable for ProverKey<T> {
...
+/// field elements per the [`Instantiable`] impl of [`AssignedAccumulator`].
+pub fn accumulator_pi_len() -> usize {
+    <AssignedAccumulator<InnerSelfEmulation> as Instantiable<outer::Scalar>>::as_public_input(
+        &Accumulator::<InnerSelfEmulation>::trivial(&[]),
+    )
+    .len()
+}
+
+/// Reconstructs a single-point-per-side accumulator from its public-input
+/// encoding (`lhs_point || lhs_scalar || rhs_point || rhs_scalar`).
```
miguel-ambrona, line 405: I would document that this returns `None` on failure (and what failures are possible).

zkir-v3/src/decider.rs
miguel-ambrona, line 1: I would avoid the name "decidable". It overlaps with the well-established term in computational complexity.

miguel-ambrona: The term "Decider" is fine though.

zkir-v3/src/ir_instructions/decidable.rs
```diff
@@ -0,0 +1,283 @@
...
+
+type AssignedPoint = <S as SelfEmulation>::AssignedPoint;
+
+/// Which deferred obligation, if any, an inner proof carries into the proof that
+/// verifies it.
+///
+/// # Wire format
+///
+/// The tag is a single byte written by declaration order: `None = 0`,
+/// `Simple = 1`. It is consensus wire format, so **never reorder or remove a
```
miguel-ambrona: Simple -> Collapsed

zswap/src/prove.rs
iquerejeta: Fixed in [08ae905](https://github.com/midnightntwrk/midnight-ledger/pull/738/commits/08ae905caf8b6d38bc14774fecbdbd3659b6686a). Note that the size increases because using `Serializable` adds a header with the size of the payload.

ledger/src/dust.rs
iquerejeta: Fixed in [08ae905](https://github.com/midnightntwrk/midnight-ledger/pull/738/commits/08ae905caf8b6d38bc14774fecbdbd3659b6686a). Note that the size increases because using `Serializable` adds a header with the size of the payload.

zkir-v3/src/ir_instructions/decidable.rs
```diff
@@ -0,0 +1,283 @@
...
+            anyhow!(
+                "a `Simple` inner proof's instance must end with the {acc_len} fields of a \
+                 collapsed accumulator, but it has {}",
+                instance.len()
+            )
+        })
+}
+
+/// Constructs the accumulator from the instance.
+fn carried_accumulator(instance: &[outer::Scalar]) -> anyhow::Result<Accumulator<S>> {
```
miguel-ambrona: This assumes the accumulator is in the tail. Document it.

zkir-v3/src/decider.rs
```diff
@@ -0,0 +1,283 @@
...
+            "the tail of a `Simple` inner proof's instance is not a well-formed collapsed \
+             accumulator; producers must normalise with `collapse -> resolve_fixed_bases -> \
+             collapse` before exposing it"
+        )
+    })?;
+
+    // The encoding is `lhs_point || lhs_scalar || rhs_point || rhs_scalar`, so
+    // each side's scalar is the last field of its half. A collapsed accumulator
+    // has both equal to one, and `assign_collapsed_accumulator` constrains them
+    // to one.
```
miguel-ambrona, line 157: If it they are always one, why do we serialize the scalars?

zkir-v3/src/ir_instructions/decidable.rs
```diff
@@ -0,0 +1,283 @@
...
+    let half = tail.len() / 2;
+    if tail[half - 1] != outer::Scalar::ONE || tail[tail.len() - 1] != outer::Scalar::ONE {
+        bail!("a carried accumulator is not collapsed: both side scalars must be one");
+    }
+
+    Ok(acc)
+}
+
+/// Final off-circuit verification step for a `DeciderKind`.
+pub fn decide_offcircuit(
```
miguel-ambrona: This is not deciding anything, it is returning an accumulator in public input format.

Given the name, I was expecting this function to check the integrity of the accumulator.

zkir-v3/src/decider.rs
```diff
@@ -0,0 +1,283 @@
...
+
+    acc.resolve_fixed_bases(bases);
+    acc.collapse();
+
+    Ok(<AssignedAccumulator<S> as Instantiable<outer::Scalar>>::as_public_input(&acc))
+}
+
+/// In-circuit mirror of [`decide_offcircuit`], constraining the single resulting
+/// accumulator as public inputs.
+///
```
miguel-ambrona, line 192: Not the mirror if this one takes "guard" and the off-circuit one does not.

zkir-v3/src/ir_instructions/decidable.rs
```diff
@@ -0,0 +1,283 @@
...
+    let verifier = std.verifier();
+    let bls = std.bls12_381();
+    let scalar_chip = bls.scalar_field_chip();
+
+    // TODO: if we use truncated challenges it may make sense to collapse before
+    // accumulating. 
+    let mut acc = match kind {
+        DeciderKind::None => own,
+        DeciderKind::Collapsed => {
+            let carried = assign_carried_accumulator(std, layouter, instance, guard)?;
```
miguel-ambrona: This should not be called "assign". I believe there is no freedom on the acc that we are building here, it must be the one corresponding to the `instance` tail.

zkir-v3/src/ir_instructions/decidable.rs
```diff
@@ -0,0 +1,283 @@
...
+
+    AssignedAccumulator::scale_by_bit(layouter, scalar_chip, guard, &mut acc)?;
+    acc.collapse(layouter, bls, scalar_chip)?;
+
+    verifier.constrain_as_public_input(layouter, &acc)
+}
+
+/// Reconstructs, in-circuit, the accumulator an inner proof carries in
+/// its instance tail.
+fn assign_carried_accumulator(
```
miguel-ambrona: Yeah, calling this "assign" is confusing.

zkir-v3/src/ir_instructions/decidable.rs
miguel-ambrona: Btw, why do you care about the "guard" here if you are then gonna "scale_by_guard" later?

zkir-v3/src/ir_instructions/decidable.rs
```diff
@@ -0,0 +1,283 @@
...
+    let assigned_accumulator = std
+        .verifier()
+        .assign_collapsed_accumulator(layouter, &[], accumulator)?;
+
+    for (selected, assigned) in selected_limbs
+        .iter()
+        .zip(std.verifier().as_public_input(layouter, &assigned_accumulator)?)
+    {
+        std.assert_equal(layouter, selected, &assigned)?;
+    }
```
miguel-ambrona: Dangerous stuff. I would make clear with comments that these 3 blocks come together.

transient-crypto/src/proofs.rs
```diff
@@ -134,13 +134,35 @@ lazy_static! {
...
 #[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Serializable, Storable)]
 #[storable(base)]
-#[tag = "proof[v5]"]
-pub struct Proof(pub Vec<u8>);
+#[tag = "proof[v6]"]
+pub struct Proof {
+    /// The raw PLONK proof bytes.
+    pub bytes: Vec<u8>,
+    /// Deferred KZG accumulators exposed by `verify_proof` instructions.
+    pub accumulators: Vec<Vec<Fr>>,
```
miguel-ambrona: Why not `Vec<Accumulator<...>>`?

transient-crypto/src/proofs.rs
```diff
@@ -397,6 +419,29 @@ pub fn reconstruct_accumulator(fields: &[outer::Scalar]) -> Option<Accumulator<I
     Some(Accumulator::new(lhs, rhs))
 }
 
+/// Rebuilds one accumulator per block of `accumulator_pi_len()` field elements
+/// carried by `proof.accumulators`.
+fn extract_accumulators(
```
miguel-ambrona: Why is there a function for this? Can't just just use `reconstruct_accumulator` from the caller?

transient-crypto/src/proofs.rs
```diff
@@ -596,22 +641,48 @@ impl VerifierKey {
...
+        // inputs (binding input, communications commitment, impact fields). The
+        // deferred KZG accumulators exposed by `verify_proof` instructions travel
+        // on the proof itself; they are prepended here to reconstruct the full
+        // public-input vector the outer circuit committed to.
+        let mut pi: Vec<outer::Scalar> = proof
+            .accumulators
+            .iter()
+            .flat_map(|block| block.iter().map(|f| f.0))
+            .collect();
+        pi.extend(statement.map(|f| f.0));
```
miguel-ambrona, line 782: This confuses me, I thought the accumulators were in the tail.

zkir-v3/src/ir_instructions/verify_proof.rs
```diff
@@ -0,0 +1,136 @@
...
+use transient_crypto::proofs::InnerSelfEmulation as S;
+
+use crate::ir_instructions::decidable::{
+    decide_incircuit, decide_offcircuit, deserialize_vk, trivial_accumulator_pis,
+};
+
+/// Label prefix for the inner verifying key's fixed bases. The specific string
+/// is arbitrary but must match between the off- and in-circuit passes so
+/// `resolve_fixed_bases` can pair each named scalar with its base.
+const VK_NAME: &str = "inner_vk";
```
miguel-ambrona: Does this mean that there can only be one inner vk?

zkir-v3/src/ir_instructions/verify_proof.rs
```diff
@@ -0,0 +1,136 @@
...
+/// If `guard` is `false` the key is not even read, and the trivial
+/// accumulator's encoding is returned instead. That is what the in-circuit pass
+/// reduces a guarded-off accumulator to, so the public inputs of the two runs
+/// agree.
+pub fn verify_proof_offcircuit(
+    vk_blob: &[u8],
+    instance: &[outer::Scalar],
+    proof: &[u8],
+    guard: bool,
+) -> anyhow::Result<Vec<outer::Scalar>> {
```
miguel-ambrona: Why this output type? Why not bool or result on unit?

If you really want to return the accumulator, why in pi form?

zkir-v3/src/ir_instructions/verify_proof.rs
```diff
@@ -0,0 +1,136 @@
...
+    )?;
+
+    let own = Accumulator::<S>::from_dual_msm(dual_msm, VK_NAME, &bases);
+    decide_offcircuit(kind, own, &bases, instance)
+}
+
+/// In-circuit mirror of [`verify_proof_offcircuit`]: verifies the inner proof
+/// in-circuit and constrains the resulting single-point accumulator as public
+/// inputs.
+///
```
miguel-ambrona, line 94: Not the mirror! The off-circuit one does not return `Result<()...>`

zkir-v3/src/ir_instructions/verify_proof.rs
```diff
@@ -0,0 +1,136 @@
...
+    let mut assigned_bases = BTreeMap::new();
+    for (name, base) in fixed_bases::<S>(VK_NAME, plonk_vk) {
+        assigned_bases.insert(name, bls.assign_fixed(layouter, base)?);
+    }
+
+    // The committed instance is a single identity point (we do not support
+    // committed instances), mirroring the off-circuit `&[&[C::identity()]]`.
+    let committed = [bls.assign_fixed(layouter, <S as SelfEmulation>::C::identity())?];
+
+    let own = verifier.prepare(layouter, &assigned_vk, &committed, instance, proof)?;
```
miguel-ambrona: How about calling this "own_acc"?

zkir-v3/src/ir.rs
```diff
@@ -47,6 +47,12 @@ pub struct IrSource {
     pub do_communications_commitment: bool,
     /// The sequence of instructions to run in-circuit
     pub instructions: Arc<Vec<Instruction>>,
+    /// Full verifying keys for the circuit's `VerifyProof` instructions. 
+    /// Each entry is
+    /// [`serialize_vk`](crate::ir_instructions::decidable::serialize_vk)'s
+    /// output: the declared `DeciderKind`'s tag byte, then the `MidnightVK`.
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub verify_proof_vks: Vec<Vec<u8>>,
```
miguel-ambrona, line 53: Well, if you can have more than one vk, then the hard-coded `VK_NAME` is wrong, it should be a different name per key.

zkir-v3/src/ir_vm.rs
```diff
@@ -642,6 +705,59 @@ impl IrSource {
...
+                            })?;
+                        proof_witnesses_idx += 1;
+                        match witness {
+                            InnerProofWitness::Direct(bytes) => bytes.clone(),
+                        }
+                    } else {
+                        // Guarded off: consume nothing, and bind the empty blob.
+                        // The `VerifyProof` under the same guard discards the
+                        // accumulator it produces from it.
+                        Vec::new()
```
miguel-ambrona: Hum, so the pool of proofs is only consumed if the guard is active? That seems like a dangerous design.

zkir-v3/src/ir_vm.rs
```diff
@@ -1168,6 +1323,57 @@ impl Relation for IrSource {
...
+                    for op in instance {
+                        let v = resolve_operand(std, layouter, &memory, op)?;
+                        let x: AssignedNative<_> = v.try_into()?;
+                        assigned_instance.push(x);
+                    }
+
+                    let proof_value = proofs
+                        .get(proof)
+                        .cloned()
+                        .ok_or_else(|| Error::Synthesis(format!("not an inner proof: {proof:?}")))?;
```
miguel-ambrona: This time the fetch of the proof is unconditional on the guard.

ledger-wasm/src/contract.rs
```diff
@@ -96,6 +98,11 @@ impl ContractCallPrototype {
                 output: from_value(output)?,
                 communication_commitment_rand: from_hex_ser(communication_commitment_rand)?,
                 key_location: KeyLocation(std::borrow::Cow::Owned(key_location.to_owned())),
+                proof_witnesses: proof_witnesses
```
miguel-ambrona: proof_witnesses -> inner_proofs

transient-crypto/src/proofs.rs
iquerejeta: These are the accumulators of the outer proof.

zkir-v3/src/ir_instructions/verify_proof.rs
iquerejeta: Adding the hash as the VK name.

zkir-v3/src/ir_instructions/decidable.rs
iquerejeta: I need the guard inside because the inner accumulator needs to have the correct structure. If the guard is false, but the field elements to represent the accumulator do not represent a valid one (i.e. valid points). It will fail. Therefore we need to use the trivial accumulator guarded behind the `guard`.

zkir-v3/src/ir_instructions/decidable.rs
iquerejeta: I've added them in a code block

zkir-v3/src/ir_instructions/decidable.rs
iquerejeta: renamed 👍

zkir-v3/src/ir_vm.rs
iquerejeta: I'm not always adding, regardless of the guard. Empty vector when it is off.

zkir-v3/src/ir_instructions/verify_proof.rs
iquerejeta: Changed the docs.

zkir-v3/src/ir_instructions/verify_proof.rs
iquerejeta: 👍  agreed. Changed the type

zkir-v3/src/decider.rs
iquerejeta: Changed the wording in the docs.

zkir-v3/src/ir_instructions/decidable.rs
iquerejeta: I'm keeping the name for consistency between the off circuit and in circuit function. Though I'm rewording the docs. 

Feel free to change the name if you feel strongly about it.

zkir-v3/src/decider.rs
iquerejeta: We encode it to use the decoder of midnight-zk.

transient-crypto/src/proofs.rs
iquerejeta: just a helper function for iterating over blocks. Its called a couple of times.

transient-crypto/src/proofs.rs
iquerejeta: I don't think we can construct one `from_public_inputs`. That function is not invertible. However, if you do find a way to create them, feel free to make the change

zkir-v3/src/ir_vm.rs
miguel-ambrona: Forget this comment, I did not realize that `Vec::new()` is actually loading a witness proof even when the guard is not active.

The verifier_gadget handles an EOF proof witness correctly. I will add a comment on this.

miguel-ambrona (dismissed):
Fine for now, but I suggest to revisit a few things:

`DeciderKind` should be provisional. It would be better that `verify_proof` returned an accumulator as a first-class IR value leaving accumulation and deferral to the circuit author. This would come with extra IR instructions for accumulating and deferring accumulators and with a Compact linter highlighting unused or undeferred accumulators.

I would personally remove the `InnerProof` instruction, it serves for loading a witness only. I would instead change the assumptions on the structure of `ProofPreimage`. This requires changes on Compact, but I recommend doing it before this feature is shipped. Let's revisit this in a couple of weeks.

Minor items that should be addressed:

- Update the fee model of verify_proof.
- Versioning issues raised by @tkerber 
- Prevent mock_verify DoS attacks
- The inner instance arity is unchecked. I suggest to expose `nb_public_inputs()` in `midnight-zk-stdlib` to be able to properly check it.

zkir-v3/src/decider.rs
miguel-ambrona: With the new type `DeferredAccumulator` we do not store the scalars.

transient-crypto/src/proofs.rs
miguel-ambrona: I introduced a new type for this. It is nice IMHO.

miguel-ambrona (dismissed):

tkerber:
Marking everything as resolved for merge criteria. After discussion, we will merge this PR under the provision that concerns will be addressed *before* the ledger 10 branch can move to stabilize to `rc`.

tkerber (dismissed):
Approval on the parts I didn't write. This will need reworking before stabilisation.

jsidorenko (approved):

PR #756, opened 2026-09-10
fix(ci): stop expanding untrusted input into workflow shell

gilescope:
Two workflows pasted `workflow_dispatch` input straight into `run:` script text. Actions expands templates before bash parses the script, so quoting inside the block cannot help: anyone with write access could inject shell.

generate-markdown-docs.yml is deleted rather than repaired. Evidence that it is dead, not merely unused:

  * 0 runs since it arrived in dde9ed55 "Initial import at 6.1.0-alpha.2"
  * it does `cd ../zswap-wasm`, and `zswap-wasm` has never existed on any ref in this repo (0 commits touching that path) - so the build job could not have succeeded on any branch
  * nothing invokes it via `workflow_call`
  * `docs/api/*` is in fact regenerated by hand in the `chore: prepare ledger-X.Y.Z` commits
  * it had drifted anyway: `zkir-wasm` and `zkir-v3-wasm` both grew a `build:markdown-docs` script and were never added

Broken did not mean safe. The injection sat on the first line of the build job's script, before the failing `cd`, and `actions/checkout` leaves GITHUB_TOKEN in `.git/config` by default - so a payload landed and could read a repo-scoped token. The advisory's `secrets.GH_PASSWORD` sites were the weaker half: they live in create-pr, which `needs: build` and therefore never ran. Reviving this pipeline should start from today's package list, not from this file.

hardfork-test-patch.yml keeps its function and gains a real guard. Its step was named "Sanitize branch name" while sanitizing nothing that mattered - the space fold ran after expansion, on text bash had already parsed. The tag now arrives as an env var, is checked against an allowlist, and is then put to `git check-ref-format` rather than restating git's ref grammar by hand. That also closes two lesser holes: a newline in the tag could forge extra step outputs through $GITHUB_OUTPUT, and `required: true` does not exclude "". Worth knowing when reviewing: this job writes GH_PASSWORD into ~/.netrc two steps earlier, so command execution there read the PAT straight off disk.

The remaining hunks in that file are the pre-existing actionlint findings the repo's pre-commit gate refuses once the file is touched at all: unquoted $HOME and five separate redirects when writing the git url rewrite (SC2086, SC2129), and two bare `*/Cargo.toml` globs (SC2035). The rewritten config block is byte-identical output, checked with `git config --get-all`.

Verified with zizmor 1.22.0: 7 high `template-injection` findings before, none after; actionlint clean on the touched file.

Not touched: upload-srs.yml:70 expands `inputs.dry-run`, but that input is
`type: boolean`, so Actions constrains it to true/false. zizmor agrees.

Assisted-by: Claude:claude-opus-5[1m] claude-code

Claude-Session: https://claude.ai/code/session_01Y7VZ4HNvVL5GnW7FBopbqM

## Sanity Checklist

This PR:
- [x] is mainly ci changes


AlexShielded (approved):
LGTM

night-warden (approved):

